PR #10, opened 2018-11-20
Google cloud pubsub

maclav3:
Work in progress


message/infrastructure/googlecloud/publisher.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	return nil
+}
+
+func (p *publisher) Close() error {
+	if p.closed {
+		return nil
+	}
+
```
roblaszczak: you should also call `topic.Stop()` on every topic - check this example: https://godoc.org/cloud.google.com/go/pubsub#Topic.Publish

message/infrastructure/googlecloud/publisher.go
```diff
@@ -0,0 +1,149 @@
...
+func (p *publisher) topic(ctx context.Context, topic string) (*pubsub.Topic, error) {
+	p.topicsLock.RLock()
+	t, ok := p.topics[topic]
+	p.topicsLock.RUnlock()
+	if ok {
+		return t, nil
+	}
+
+	p.topicsLock.Lock()
```
roblaszczak: maybe we should do lock per topic to not lock another topics?

message/infrastructure/googlecloud/subscriber.go
```diff
@@ -0,0 +1,221 @@
...
+	unmarshaler Unmarshaler
+	logger      watermill.LoggerAdapter
+}
+
+type SubscriberConfig struct {
+	SubscriptionName            string
+	ProjectID                   string
+	CreateSubscriptionIfMissing bool
+
+	ConsumersCount int
```
roblaszczak: ConsumersCount is not used anywhere (and is probably not needed?)

message/infrastructure/googlecloud/subscriber.go
```diff
@@ -0,0 +1,221 @@
...
+	config SubscriberConfig
+
+	unmarshaler Unmarshaler
+	logger      watermill.LoggerAdapter
+}
+
+type SubscriberConfig struct {
+	SubscriptionName            string
+	ProjectID                   string
+	CreateSubscriptionIfMissing bool
```
roblaszczak: also, sounds like redundant (we always want to do this?)

message/infrastructure/googlecloud/subscriber.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	err = sub.Receive(ctx, func(ctx context.Context, pubsubMsg *pubsub.Message) {
+		msg, err := s.unmarshaler.Unmarshal(pubsubMsg)
+		if err != nil {
+			s.logger.Error("could not unmarshal Google Cloud PubSub message", err, nil)
+			return
+		}
+
+		output <- msg
```
roblaszczak: it should be 

```go
select {
    case output <- msg:
        // ok
     case <-s.closing:
         return errors.New("some error")
}
```
because when client is not consuming and we will call Close it will block forever

roblaszczak: you also should check that when returning error it will cause nack

message/infrastructure/googlecloud/subscriber.go
```diff
@@ -0,0 +1,221 @@
...
+const subscriptionIDTemplate = "watermill_%s_%d"
+
+// subscription obtains a subscription object. If subscription doesn't exist on PubSub, create it.
+// subsequent calls to `subscription` with the same `topic` return separate subscriptions,
+// with ids according to `subscriptionIDTemplate`.
+func (s *subscriber) subscription(ctx context.Context, topic string) (sub *pubsub.Subscription, err error) {
+	s.activeSubscriptionsLock.RLock()
+	var subscriptionID string
+	for i := 0; ; i++ {
+		subscriptionID = fmt.Sprintf(subscriptionIDTemplate, topic, i)
```
roblaszczak: won't it be easier to use map[string][int]?

roblaszczak: because this for is hmm, strange :poultry_leg:

message/infrastructure/googlecloud/subscriber.go
```diff
@@ -0,0 +1,221 @@
...
+
+	defer func() {
+		if err != nil {
+			s.activeSubscriptionsLock.Lock()
+			s.activeSubscriptions[sub.ID()] = struct{}{}
+			s.activeSubscriptionsLock.Unlock()
+		}
+	}()
+
+	sub = s.client.Subscription(subscriptionID)
```
roblaszczak: isn't subscriptionID something like consumer group?

roblaszczak: if yes, it should be probably injected to the constructor

roblaszczak: and another question: does Google Cloud Pub/Sub work in this way:
- I create some topic
- I publish some messages to the topic
- I create consumer and consume produced messages from the topic

or:
- I create some topic
- I publish some messages to the topic
- I create consumer and there is nothing to consume on the topic
- I publish some messages to the topic
- I consume messages with 

or:
- I create some topic
- I publish some messages to the topic
- I create consumer with some subscriptionID and consume produced messages from the topic
- I create consumer with another subscriptionID and also consume produced messages from the topic

daigo: The official document would help: https://cloud.google.com/pubsub/docs/subscriber

>  Only messages published to the topic after the subscription is created are available to subscriber applications.

message/infrastructure/googlecloud/publisher.go
maclav3: 75e400b68c7aac4ca61b59c0b03fa16c8c54136d

message/infrastructure/googlecloud/publisher.go
maclav3: That would require a thread-safe map I guess. So we would end up with a similar situation as now. I could use https://golang.org/pkg/sync/#Map, but it's not pretty as we would need to convert `interface{}`.

Anyway, `topicsLock` is a `RWMutex`, so it `Lock`s only when the function is called with a new `topic`, otherwise it's a `RLock`. So it doesn't hurt that much?

message/infrastructure/googlecloud/subscriber.go
maclav3: True, it's a leftover from the early prototype. The google client library takes care of spawning goroutines itself.

message/infrastructure/googlecloud/publisher.go
roblaszczak: It may hurt a bit when for some reason we are creating a lot of topics.

message/infrastructure/googlecloud/subscriber.go
maclav3: a183611c6d4e3116d5cd766745f072da6d08f2f3

message/infrastructure/googlecloud/subscriber.go
maclav3: I wonder which case is more frequent: 
1) I am calling `Subscribe` and don't want to care whether there is a subscription or not,
or
2) I only want to use preexisting subscriptions and receive an error if no such subscription is present.

I suppose 1) is way more prevalent. I cannot think of a case where I would want 2), but I guess it's possible. This could be left as a configuration detail, but:

a) Like you say, the default would probably need to be 1), as this seems to be the more natural behavior,
b) It may be difficult to convey this configuration option clearly, so that it's self-explanatory and doesn't pollute the config.

roblaszczak: lgtm

message/infrastructure/googlecloud/subscriber.go
maclav3: The callback to `subscription.Receive` doesn't return any error. 

Anyway, when `s.closing` receives, `ctx` is canceled before `output` is closed. So with luck, we should never get a write on a closed `output` channel, because `Receive` is canceled with `ctx`.

If a race happens and `output` is closed before `Receive` is effectively canceled, writing to a closed channel would cause a panic, which may actually make sense here, as long as it is recovered in a sane way. Because it kind of is an abnormal situation.

So if there's no error to return here, maybe I should just recover from panic with an error which causes Nack? Seems a bit artificial though.

maclav3: OK, now I understand what you meant - that there would be no reader for `output` on the other end and it would hang.

2df32f9f8948f714c0bb0705fd609119423a8c5a should fix this, and pass `Ack/Nack` back to the client library, as specified in https://godoc.org/cloud.google.com/go/pubsub#Message.Ack

I'm not entirely sure if I'm using `Nack` correctly. https://godoc.org/cloud.google.com/go/pubsub suggests that sending a `Nack` tells the server "could not process this message, please redeliver ASAP". But if the message is, for example, malformed, it would make no sense to redeliver, rather Ack and put it on a poison queue, I guess?

message/infrastructure/googlecloud/subscriber.go
maclav3: It actually was pretty weird :laughing: 
532ef915c1ca7188e48f9639082219c9c7d2054b

message/infrastructure/googlecloud/subscriber.go
maclav3: It is somewhat similar. See 532ef915c1ca7188e48f9639082219c9c7d2054b, now it is taken from the config. I will add validation to make sure that it is not left empty.

message/infrastructure/googlecloud/subscriber.go
roblaszczak: > "could not process this message, please redeliver ASAP"
this is good, when message is malformed poision queue of retry should be used, but it is not responsibility of the PubSub

PR #25, opened 2018-12-18
Improve website wording

sagikazarmark:
I think it's less likely to misunderstand this way. :)

Noticed by @orymate


roblaszczak:
Make sense, thanks :)

PR #27, opened 2019-01-05
HTTP publisher

maclav3:


message/infrastructure/http/pubsub_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/ThreeDotsLabs/watermill/message"
+	"github.com/ThreeDotsLabs/watermill/message/infrastructure"
+)
+
+func createPubSub(t *testing.T) message.PubSub {
+	logger := watermill.NewStdLogger(true, true)
+
+	sub, err := NewSubscriber(":8080", DefaultUnmarshalMessageFunc, logger)
```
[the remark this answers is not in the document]
maclav3: From what I've read, `ListenAndServe` is blocking and there is no way to check when it's ready. The `Sleep` was ugly and hopefully, temporary. [Here](https://grokbase.com/t/gg/golang-nuts/141ee4dqyg/go-nuts-how-to-know-when-listenandserve-is-ready-to-handle-connections#201401153vyc35xvn3kqhbzrfvg2rqdivy) they say that you can call `Listen("tcp", addr) and then `go http.Serve()` and be sure that the server is already listening. But `Subscriber` doesn't expose `Listen`, only `ListenAndServe`. However, I don't feel too good about exposing `Listen` as well, because it's a bit low-level. Maybe I could rewrite `Subscriber.ListenAndServe` to use this hack.

message/infrastructure/http/pubsub_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	go func() {
+		err := sub.StartHTTPServer()
+		require.NoError(t, err)
+	}()
+
+	time.Sleep(200 * time.Millisecond)
```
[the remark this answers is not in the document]
maclav3: From what I've read, `ListenAndServe` is blocking and there is no way to check when it's ready. The `Sleep` was ugly and hopefully, temporary. [Here](https://grokbase.com/t/gg/golang-nuts/141ee4dqyg/go-nuts-how-to-know-when-listenandserve-is-ready-to-handle-connections#201401153vyc35xvn3kqhbzrfvg2rqdivy) they say that you can call `Listen("tcp", addr)` and then `go http.Serve()` and be sure that the server is already listening. But `Subscriber` doesn't expose `Listen`, only `ListenAndServe`. However, I don't feel too good about exposing `Listen` as well, because it's a bit low-level. Maybe I could rewrite `Subscriber.ListenAndServe` to use this hack.

message/infrastructure/http/metadata.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/ThreeDotsLabs/watermill/message"
+)
+
+type messageMetadata struct {
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
+func metadataToJson(msg *message.Message) ([]byte, error) {
+	metadataItems := make([]messageMetadata, len(msg.Metadata))
```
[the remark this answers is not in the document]
maclav3: 4dc86cc4ce23a0b44c62d970f29c39a71437fdfb
 Implemented json.Marshaler and json.Unmarshaler directly on message.Metadata and removed `metadata.go`

message/infrastructure/http/publisher.go
```diff
@@ -0,0 +1,104 @@
...
+	marshalMessageFunc MarshalMessageFunc
+
+	closed bool
+}
+
+func NewPublisher(marshalMessageFunc MarshalMessageFunc, logger watermill.LoggerAdapter) (*Publisher, error) {
+	return NewPublisherWithClient(http.DefaultClient, marshalMessageFunc, logger)
+}
+
+func NewPublisherWithClient(client *http.Client, marshalMessageFunc MarshalMessageFunc, logger watermill.LoggerAdapter) (*Publisher, error) {
```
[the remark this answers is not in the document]
maclav3: If `DefaultMarshalMessageFunc` is not included in the default config, perhaps we should think of another name for it. Because it's not default then.

How about `HTTPPostRequestMarshalFunc` or something similar?

message/infrastructure/http/pubsub_test.go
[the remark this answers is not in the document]
maclav3: This could probably be achieved with the marshaling function, because it sets the request URL anyway. PoC coming soon.

message/infrastructure/http/subscriber.go
```diff
@@ -1,16 +1,45 @@
...
 	"github.com/ThreeDotsLabs/watermill"
 	"github.com/ThreeDotsLabs/watermill/message"
 	"github.com/go-chi/chi"
 )
 
 type UnmarshalMessageFunc func(topic string, request *http.Request) (*message.Message, error)
 
+// DefaultUnmarshalMessageFunc retrieves the UUID and Metadata from request headers,
+// as encoded by DefaultMarshalMessageFunc.
+func DefaultUnmarshalMessageFunc(topic string, req *http.Request) (*message.Message, error) {
```
roblaszczak: also, some unit tests for DefaultMarshalMessageFunc will be great

message/infrastructure/http/publisher.go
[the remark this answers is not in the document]
roblaszczak: Well, it will be still something "default" but not automatically set to the config. Eventually we can set it to the config by default, but in this case, we need to not forget to explicitly define this function in docs, etc. in other case users may not easily notice how to replace the default implementation

message/infrastructure/http/pubsub_test.go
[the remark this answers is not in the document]
roblaszczak: but isn't the http client himself waiting for the server when it's not available?

message/infrastructure/http/publisher.go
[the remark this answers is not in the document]
maclav3: afc279656462578257cb05721e69ef1835d5c3e3 adds configs for both Pub and Sub

message/infrastructure/http/publisher.go
```diff
@@ -0,0 +1,104 @@
...
+			"method":   req.Method,
+			"provider": ProviderName,
+		}
+
+		resp, err := p.client.Do(req)
+		if err != nil {
+			return errors.Wrapf(err, "publishing message %s failed", msg.UUID)
+		}
+
+		// todo: process the response anyhow?
```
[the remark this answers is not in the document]
maclav3: a9666ef7fec40fcf8c6a774ae04a87e059e46caa ok, now I'm quite satisfied with how errors are handled from this request. Also, I made sure that `resp.Body.Close()` is always called to prevent leaks, even when there was no server error.

message/infrastructure/http/pubsub_test.go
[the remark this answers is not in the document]
maclav3: I've had the test fail because of a race condition. I've removed the sleep already, but i've had to modify `ListenAndServe` (`Listen(), go Serve, return`) to prevent a case where there would be nowhere to send the request to.

message/infrastructure/http/publisher.go
```diff
@@ -0,0 +1,155 @@
...
+		p.handleResponseBody(resp, logFields)
+		if resp.StatusCode >= http.StatusBadRequest {
+			return errors.Wrapf(err, "%d %s", resp.StatusCode, resp.Status)
+		}
+
+		if err != nil {
+			return errors.Wrapf(err, "could not close response body for message %s", msg.UUID)
+		}
+
+		p.logger.Trace("message published", logFields)
```
roblaszczak: cosmetics, but logs are starts from capital letter :poultry_leg:

message/infrastructure/http/publisher.go
```diff
@@ -0,0 +1,155 @@
...
+			"method":   req.Method,
+			"provider": ProviderName,
+		}
+
+		resp, err := p.config.Client.Do(req)
+		if err != nil {
+			return errors.Wrapf(err, "publishing message %s failed", msg.UUID)
+		}
+
+		p.handleResponseBody(resp, logFields)
```
roblaszczak: err not handled

message/infrastructure/http/pubsub_test.go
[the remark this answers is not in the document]
maclav3: 61cf19975cf1496710cd87e4319100e5bacefab3 `:0` port actually works. Every parallel test has a different port

message/infrastructure/http/subscriber.go
```diff
@@ -110,8 +141,24 @@ func (s *Subscriber) Subscribe(url string) (chan *message.Message, error) {
 
 // StartHTTPServer starts http server.
 // StartHTTPServer must be called after all subscribe function are called.
-func (s *Subscriber) StartHTTPServer() error {
-	return s.server.ListenAndServe()
+func (s *Subscriber) StartHTTPServer() (chan error, error) {
```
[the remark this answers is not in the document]
maclav3: If we want this method to be non-blocking and guarantee that the listener is ready after it returns, I guess we have to return a `chan error` for the eventual error when `Serve()` ends and `error` for the immediate error when creating the listener fails.

I've had the HTTP test fail earlier, because the subscriber server wouldn't be ready when the publisher tried to connect to it, as `ListenAndServe` provides no way to wait for the listener to be up.

message/infrastructure/http/publisher.go
maclav3: 6514a6e07ebd4511dd82edda102cbe54314d09c1

message/infrastructure/http/publisher.go
maclav3: a75e6e20b83469918e391a07de27e09b66fd3c2f

docs/content/docs/pub-sub-implementations.md
```diff
@@ -146,12 +146,10 @@ marshaler := kafka.NewWithPartitioningMarshaler(func(topic string, msg *message.
 
 ### HTTP
 
-At this moment only HTTP subscriber is available. There is an issue for a [HTTP publisher](https://github.com/ThreeDotsLabs/watermill/issues/17).
-
-HTTP subscriber allows us to send messages received by HTTP request (for example - webhooks).
+The HTTP subscriber listens to HTTP requests (for example - webhooks) and outputs them as messages.
```
roblaszczak: `  [HTTP]({{< ref "#http" >}})  |   | x | `prod-ready` |` should be changed at the bottom of the page (add publisher)

docs/content/docs/pub-sub-implementations.md
```diff
@@ -146,12 +146,10 @@ marshaler := kafka.NewWithPartitioningMarshaler(func(topic string, msg *message.
...
 You can then post them to any Publisher. Here is an example with [sending HTTP messages to Kafka](https://github.com/ThreeDotsLabs/watermill/blob/master/_examples/http-to-kafka/main.go).
 
-When implemented, HTTP publisher can be used as webhooks sender.
+The HTTP publisher sends HTTP requests as specified in its configuration. Here is an example with [transforming Kafka messages into HTTP webhook requests](https://github.com/ThreeDotsLabs/watermill/tree/http-publisher/_examples/kafka-to-http).
```
roblaszczak: link will be broken after merge and branch remove

docs/content/docs/pub-sub-implementations.md
```diff
@@ -162,21 +160,34 @@ When implemented, HTTP publisher can be used as webhooks sender.
 | GuaranteedOrder | yes |  |
 | Persistent | no| |
 
-#### Configuration
+#### Subscriber configuration
```
roblaszczak: ExactlyOnceDelivery should be true:)

message/infrastructure/http/publisher.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+type MarshalMessageFunc func(topic string, msg *message.Message) (*http.Request, error)
+
+// DefaultMarshalMessageFunc returns a MarshalMessage func transforming the message into a HTTP POST request.
+// It encodes the UUID and Metadata in request headers.
+// The request URL is combined from the base address and the topic.
+func DefaultMarshalMessageFunc(address string) MarshalMessageFunc {
+	return func(topic string, msg *message.Message) (*http.Request, error) {
+		req, err := http.NewRequest(http.MethodPost, address+"/"+topic, bytes.NewBuffer(msg.Payload))
```
roblaszczak: what about just changing topic to entire url?

docs/content/docs/pub-sub-implementations.md
```diff
@@ -162,21 +160,34 @@ When implemented, HTTP publisher can be used as webhooks sender.
...
+
+Publisher configuration is done via the config struct passed to the constructor:
+
 {{% render-md %}}
-{{% load-snippet-partial file="content/src-link/message/infrastructure/http/subscriber.go" first_line_contains="// NewSubscriberWithRouter" last_line_contains="(*Subscriber, error)" %}}
+{{% load-snippet-partial file="content/src-link/message/infrastructure/http/publisher.go" first_line_contains="type PublisherConfig struct" last_line_contains="}" %}}
 {{% /render-md %}}
 
+How the message topic and body translate into the URL, method and payload of the HTTP request is highly configurable through the use of `MarshalMessageFunc`. 
+Use the provided `DefaultMarshalMessageFunc` to send POST requests to a predefined address:
```
roblaszczak: some tip that you probably want to implement your own marshal function will be useful

_examples/http-producer/main.go
```diff
@@ -0,0 +1,141 @@
+package main
```
roblaszczak: it is outdated? if not that some readme will be useful

_examples/kafka-to-http/README.md
```diff
@@ -0,0 +1,32 @@
+# HTTP producer example
+
+This example showcases the use of HTTP producer to call webhooks with HTTP POST requests. It consists of three services:
+
+1. `producer` publishes messages on Kafka. The messages come in three varieties: `Foo`, `Bar`, and `Baz`. The event type is encoded in metadata, under the key `event_type`.
+1. `webhook_server` is a HTTP server that listens for requests and prints the path, method, and payload on stdout.
```
roblaszczak: it is inconsisten with the directorny name (`http-server`), it's good idea to unify it

_examples/kafka-to-http/router/main.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+var (
+	brokers   = []string{"kafka:9092"}
+	logger    = watermill.NewStdLogger(false, false)
+	topicName = "kafka_to_http_example"
+)
+
+// passMessages passes the message along if its event type is one of acceptedTypes.
+func passMessages(acceptedTypes ...string) message.HandlerFunc {
```
roblaszczak: maybe `filterMessages` will be better name?

_examples/kafka-to-http/router/main.go
```diff
@@ -0,0 +1,69 @@
...
+	return func(msg *message.Message) ([]*message.Message, error) {
+		msgEventType := msg.Metadata.Get("event_type")
```
roblaszczak: some comment, that metadata is set in producer, and we avoiding unmarshaling entire event to check event type when saved to metadata

message/infrastructure/http/retry.go
```diff
@@ -0,0 +1,83 @@
...
+
+	return nil
+}
+
+type RetryPublisher struct {
+	pub    *Publisher
+	config RetryPublisherConfig
+}
+
+func NewRetryPublisher(pub *Publisher, config RetryPublisherConfig) (*RetryPublisher, error) {
```
roblaszczak: it's not documented here and in watermill.io

message/infrastructure/http/retry.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, errors.Wrap(err, "invalid RetryPublisher config")
+	}
+
+	return &RetryPublisher{
+		pub,
+		config,
+	}, nil
+}
+
+func (p RetryPublisher) Publish(topic string, messages ...*message.Message) error {
```
roblaszczak: it is pretty simple to test (and to broke :D), so you know :poultry_leg:

message/infrastructure/http/retry.go
roblaszczak: and in theory it is not limited to http, maybe its' good idea to move it out of http package?

message/infrastructure/http/retry.go
```diff
@@ -0,0 +1,83 @@
...
+	var err error
+	timeToNextRetry := p.config.TimeToFirstRetry
+
+	for i := 0; i < p.config.MaxRetries; i++ {
+		err = p.pub.Publish(topic, messages...)
```
roblaszczak: that hurts me a bit, that all messages are resent

message/infrastructure/http/subscriber.go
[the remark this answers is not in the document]
roblaszczak: i'm not sure that breaking API for tests to pass is an good idea, and previous API was easier (channels are always problems), maybe for tests we would just wait for open port istead?

message/metadata.go
```diff
@@ -1,7 +1,14 @@
 package message
 
+import "encoding/json"
+
 type Metadata map[string]string
 
+type metadataItem struct {
```
roblaszczak: but why not to just use native map marshal?

message/infrastructure/http/subscriber.go
maclav3: 28107de858177132c98e27adc2100338411e459c, updated in b97952c46ffe87d5349ecb9ef5a191baa72fbd80 (changed signature)

message/infrastructure/http/subscriber.go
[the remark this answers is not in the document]
maclav3: 907188dc4d91e3019e927e0f775727c011289c7f

docs/content/docs/pub-sub-implementations.md
maclav3: 988ec6531abbd023c5fcba55b153e3c8ba953f13

docs/content/docs/pub-sub-implementations.md
maclav3: 988ec6531abbd023c5fcba55b153e3c8ba953f13

docs/content/docs/pub-sub-implementations.md
maclav3: 988ec6531abbd023c5fcba55b153e3c8ba953f13

message/infrastructure/http/publisher.go
maclav3: b97952c46ffe87d5349ecb9ef5a191baa72fbd80

_examples/http-producer/main.go
maclav3: It's outdated, I've removed it altogether

_examples/kafka-to-http/router/main.go
maclav3: ef529080bec252977f397f29a8098a4bbfc6222b

_examples/kafka-to-http/router/main.go
maclav3: ef529080bec252977f397f29a8098a4bbfc6222b

message/infrastructure/http/retry.go
maclav3: 2250244364ee79e11de27e1cf0ada0db189fa4d4

message/infrastructure/http/retry.go
maclav3: 2250244364ee79e11de27e1cf0ada0db189fa4d4

message/metadata.go
maclav3: yes 4944c4db8425df59013ee6bc96ebf7d6fcfbad36

message/infrastructure/http/retry.go
maclav3: Yep, quite ugly. However, the Publisher interface provides no easy way to check which messages were successfully sent, so I've birthed this monstrosity: 7a8a0ce1c4bad2e5801af53baf1610023f007e6d

It takes each message and repeats it. Errors are gathered in a composite error type. The unit tests are somewhat contrived, but cover the actual problems I had while writing this

docs/content/docs/pub-sub-implementations.md
maclav3: umm that's quite obvious, I guess

internal/publisher/retry.go
```diff
@@ -0,0 +1,105 @@
...
+// send sends one message at a time to prevent sending a successful message more than once.
+func (p RetryPublisher) send(topic string, msg *message.Message) error {
+	var err error
+	timeToNextRetry := p.config.TimeToFirstRetry
+
+	for i := 0; i < p.config.MaxRetries; i++ {
+		err = p.pub.Publish(topic, msg)
+		if err == nil {
+			return nil
+		}
```
roblaszczak: it will be an a good idea to log the error (info level is probably enough)

_examples/kafka-to-http/router/main.go
```diff
@@ -0,0 +1,69 @@
...
+			if string(typ) == msgEventType {
+				return message.Messages{msg}, nil
+			}
+		}
+
+		return nil, nil
+	}
+}
+
+func marshalMessage(topic string, msg *message.Message) (*http.Request, error) {
```
roblaszczak: what you think about just removing this function, using DefaultMarshalMessageFunc and url in `router.AddHandler()`, it will be much more clear

_examples/kafka-to-http/router/main.go
```diff
@@ -0,0 +1,69 @@
...
+	subscriber, err := kafka.NewSubscriber(kafka.SubscriberConfig{
+		Brokers: brokers,
+	}, nil, kafka.DefaultMarshaler{}, logger)
+	if err != nil {
+		panic(err)
+	}
+
+	pubSub := message.NewPubSub(publisher, subscriber)
+
+	router, err := message.NewRouter(message.RouterConfig{CloseTimeout: 5 * time.Second}, logger)
```
roblaszczak: you can consider to remove `CloseTimeout: 5 * time.Second` to simplify example

message/infrastructure/googlecloud/subscriber.go
```diff
@@ -218,6 +218,10 @@ func (s *Subscriber) receive(
 			return
 		}
 
+		msgCtx, cancel := context.WithCancel(ctx)
```
roblaszczak: it's interesting why tests were passing, did you investigated it?

message/infrastructure/http/subscriber.go
```diff
@@ -1,23 +1,69 @@
...
+// DefaultUnmarshalMessageFunc retrieves the UUID and Metadata from request headers,
+// as encoded by DefaultMarshalMessageFunc.
+func DefaultUnmarshalMessageFunc(topic string, req *http.Request) (*message.Message, error) {
+	body, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	uuid := req.Header.Get(HeaderUUID)
+	if uuid == "" {
```
roblaszczak: it's imo fine, uuid is not required anywhere

roblaszczak: and it may be hard to enforce every client remote publisher to send this header (for example: gitlab)

message/infrastructure/http/subscriber.go
```diff
@@ -1,23 +1,69 @@
...
+		return nil, errors.New("No UUID encoded in header")
+	}
+
+	msg := message.NewMessage(uuid, body)
+
+	metadata := req.Header.Get(HeaderMetadata)
```
roblaszczak: like above, probably we should bypass it when the header is empty

_examples/kafka-to-http/router/main.go
maclav3: 82f0556f36e29d8788da5984ab4a37909ac5e8c2

_examples/kafka-to-http/router/main.go
maclav3: 82f0556f36e29d8788da5984ab4a37909ac5e8c2

message/infrastructure/googlecloud/subscriber.go
maclav3: IIRC, they weren't. Nobody checked since.
Tests in CI recommended :)

internal/publisher/retry.go
maclav3: 00fb68ea15f020c05e491287adf558a730bb32af

message/infrastructure/http/subscriber.go
maclav3: 744769805d87476f08dc58de6626dfc38ae98125

message/infrastructure/http/subscriber.go
maclav3: 744769805d87476f08dc58de6626dfc38ae98125

PR #28, opened 2019-01-10
Set Kafka message key

slavirok:
Kafka message key was not set previously. Because of that, I was not able to publish messages to Compact Kafka topics.


roblaszczak:
This can be already achieved by using `kafkaJsonWithPartitioning`. Please check: https://watermill.io/docs/pub-sub-implementations/#partitioning

Anyway, thanks for PR :)

PR #35, opened 2019-02-04
Replace Travis with GitLab CI

m110:


.ci/wait-for-services.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -e
+
+readonly WAIT_FOR_IT_URL=https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh
```
roblaszczak: maybe it is better to make hard copy to repo?

m110: :+1:

PR #38, opened 2019-02-05
Use latest patch release of Go 1.11

AlekSi:


roblaszczak:
Good idea, thanks! :1st_place_medal:

PR #50, opened 2019-02-15
Proposal to enable race detector in tests

cristaloleg:
Unless you've objections for this flag 😉


m110:
@cristaloleg thanks for pointing that out. We don't have objections for the flag, although the main tests suite is pretty advanced and spawns a lot of goroutines. We had some issues running these with the race detector, as it timed out.

Our proposal is to run the "short" tests with `-race` enabled on master branch, along with the full tests suite. For commits on other branches, we'd like to stick to short tests.

I've created a PR with these changes and I'd appreciate your feedback: #51

cristaloleg:
Thanks @m110 for the detailed explanation.
Your PR does it better.
Done in #51

PR #70, opened 2019-05-07
make routing key dynamic in queue bind

bkielbasa:
## The problem
We use the routing key in bindings.
image

The routing key is dynamic and in our example is an event name. Because the routing key is static, we don't have the flexibility we need.

In this PR, I change the variable to a function similar to routing keys in PublishConfig etc

### DIFF
--- a/message/infrastructure/amqp/config.go
+++ b/message/infrastructure/amqp/config.go
@@ -42,7 +42,11 @@ func NewDurablePubSubConfig(amqpURI string, generateQueueName QueueNameGenerator
 			GenerateName: generateQueueName,
 			Durable:      true,
 		},
-		QueueBind: QueueBindConfig{},
+		QueueBind: QueueBindConfig{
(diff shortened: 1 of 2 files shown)


sagikazarmark:
Would be nice to have a static and a non-static version, where the function takes precedence.

roblaszczak:
@sagikazarmark you can always use 

```
func(topic string) string {
	return "something_static"
}
```

we have already pattern like this in some places and it is pretty good because it is flexible and simple (if it will be a static option we need to always remember to call both, with adds complexity)

roblaszczak:
@bkielbasa thanks 🍻

sagikazarmark:
@roblaszczak Yeah, I know, but I still think it would be nice to have a shorthand for that syntax.

alfiesal:
It is possible to bind the queue to exchange by multiple routing keys?

PR #76, opened 2019-05-17
Add interface and define interface for configuration of infrastructure component

MartinForReal:
resolve issue: #77
dd interface and define an interface for configuration of the infrastructure component

### DIFF
--- a/message/infrastructure/interface.go
+++ b/message/infrastructure/interface.go
@@ -1,0 +1,8 @@
+package infrastructure
+
+import "flag"
+
+type Config interface {
+	GetFlagSet(*flag.FlagSet)
+	Validate() error
+}


MartinForReal:
```
func (config *Config) GetFlags() *pflag.FlagSet {
    flagSet := pflag.NewFlagSet(“server”, pflag.ContinueOnError)
    flagSet.AddFlagSet(config.Cluster.GetFlags())
    flagSet.AddFlagSet(config.EventBucketConfig.GetFlags())
    flagSet.AddFlagSet(config.TaskBucketConfig.GetFlags())
    flagSet.AddFlagSet(config.ResourceBucketConfig.GetFlags())
    flagSet.AddFlagSet(config.GrpcConfig.GetFlags())
    return flagSet
}
```

roblaszczak:
As I said in Slack

> but in general I think that they are a looot of configuration methods, and it's hard to define something which is universal and it should be up to user to define it

In Watermill (and in idiomatic Go, I suppose ;) ) we are aiming to provide only generic solutions which can be used by anyone and it is hard to make some generic configuration. In theory, we can use a package like spf13/viper, but not everyone may be happy with using this solution and they may want to use another.  Because of it, it should be up to the user to load configuration by preferred method. 

I hope that it sounds reasonable for you :wink:

PR #78, opened 2019-05-20
Add CQRS marshalers

sagikazarmark:


components/cqrs/marshaler_name.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (m *NameMarshaler) Name(v interface{}) string {
+	if v, ok := v.(namedMessage); ok {
+		return v.Name()
+	}
+
+	return m.marshaler().Name(v)
+}
+
+func (m *NameMarshaler) NameFromMessage(msg *message.Message) string {
```
bkielbasa: hmm, I'm not sure if the marshaller should have this method. Maybe I'm losing some context here but for me, it should go to `message.Message`.

sagikazarmark: It's part of the `CommandEventMarshaler` interface

components/cqrs/marshaler_structname.go
```diff
@@ -0,0 +1,49 @@
...
+
+	return msg, nil
+}
+
+func (m *StructNameMarshaler) Unmarshal(msg *message.Message, v interface{}) error {
+	return m.marshaler().Unmarshal(msg, v)
+}
+
+func (m *StructNameMarshaler) Name(v interface{}) string {
+	segments := strings.Split(fmt.Sprintf("%T", v), ".")
```
roblaszczak: if v is pointer it will have `*` prefix, do we want to remove it? :wink:

sagikazarmark: I think so, but it gets removed: https://play.golang.org/p/97dR-e246vB

components/cqrs/marshaler_structname.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/ThreeDotsLabs/watermill/message"
+)
+
+// StructNameMarshaler uses the struct name (without the package) as the message name.
+//
+// It wraps another marshaler which takes care of the message serialization/deserialization.
+// If none provided, it falls back to JSON marshaler.
+type StructNameMarshaler struct {
+	Marshaler CommandEventMarshaler
```
roblaszczak: after some thinking, maybe it will be a better idea to not add it as an implementation of Marshaler interface, but like function and add an extra attribute for existing marshallers, like:

```
type JSONMarshaler struct {
	NewUUID func() string
        GenerateName func(v interface{}) string
}
```
with default `GenerateName = ObjectName`.
In that case, you may rename `object.go` to something more meaningful, and keep all names generations methods like functions

In that case, the implementations are still more about output format, and naming is decoupled.

What do you think? :)

sagikazarmark: Sounds good to me. I will update the PR soon

components/cqrs/name.go
```diff
@@ -0,0 +1,42 @@
...
+func ObjectName(v interface{}) string {
+	s := fmt.Sprintf("%T", v)
+	s = strings.TrimLeft(s, "*")
+
+	return s
+}
+
+// StructName name returns struct name in format [type name].
+// It ignores if the value is a pointer or not.
+func StructName(v interface{}) string {
```
roblaszczak: maybe ObjectNameWithoutPackage? it is pretty long, but explicit :)

components/cqrs/name.go
```diff
@@ -0,0 +1,42 @@
...
+	segments := strings.Split(fmt.Sprintf("%T", v), ".")
+
+	return segments[len(segments)-1]
+}
+
+type namedMessage interface {
```
roblaszczak: maybe just Named and public?

components/cqrs/name.go
```diff
@@ -0,0 +1,42 @@
...
+	Name() string
+}
+
+// MessageName returns the name from a message implementing the following interface:
+// 		type namedMessage interface {
+// 			Name() string
+// 		}
+// It ignores if the value is a pointer or not.
+func MessageName(fallback func(v interface{}) string) func(v interface{}) string {
```
roblaszczak: what do you think about NamedObject?

components/cqrs/marshaler_structname.go
roblaszczak: ah, of course it is :)

components/cqrs/name.go
sagikazarmark: It does not have to be public as there is no reference to it at all outside of the function.

components/cqrs/name.go
sagikazarmark: I was rather thinking about renaming `ObjectName` to `FullyQualifiedStructName` and leave this as is, but `ObjectNameWithoutPackage` works as well.

components/cqrs/name.go
sagikazarmark: Following the pattern I suggested above, this could be `NamedStruct`, but if we stick with object, `NamedObject` works.

components/cqrs/name.go
roblaszczak: :+1:

components/cqrs/name.go
roblaszczak: ok, like I like the idea of FullyQualifiedStructName :)

components/cqrs/name.go
roblaszczak: :+1:

sagikazarmark:
🎉 Thanks @roblaszczak

PR #82, opened 2019-06-01
changed AddNoPublisherHandler signature

roblaszczak:


message/router.go
```diff
@@ -30,6 +30,9 @@ var (
 // (because msg.Ack() was sent in HandlerFunc or Subscriber supports multiple consumers).
 type HandlerFunc func(msg *Message) ([]*Message, error)
 
+// NoPublishHandlerFunc is HandlerFunc alternative, which doesn't produce any message.
```
m110:
```suggestion
// NoPublishHandlerFunc is HandlerFunc alternative, which doesn't produce any messages.
```

message/router.go
```diff
@@ -30,6 +30,9 @@ var (
...
+type NoPublishHandlerFunc func(msg *Message) error
```
m110: How about `ReadOnlyHandlerFunc` or `ConsumeOnlyHandlerFunc`?

roblaszczak: sounds better @maclav3 ? and does we want to also change AddNoPublisherHandler?

PR #84, opened 2019-06-01
Add context to router.Run() and propagate it to all subscribers

m110:


message/router_test.go
```diff
@@ -310,7 +309,7 @@ func TestRouterNoPublisherHandler(t *testing.T) {
 		},
 	)
 
-	go r.Run()
+	go r.Run(context.Background())
```
golangcibot: Error return value of `r.Run` is not checked (from `errcheck`)

PR #86, opened 2019-06-01
added context cancelling support

roblaszczak:


message/router/middleware/retry.go
```diff
@@ -28,25 +26,39 @@ type Retry struct {
...
+		startTime := time.Now()
 
 		for {
 			events, err := h(message)
-			if r.shouldRetry(err, retries) {
+			if err == nil {
+				return events, nil
+			}
+
+			elapsedTime := time.Now().Sub(startTime)
```
golangcibot: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

message/router/middleware/retry_test.go
```diff
@@ -91,3 +92,84 @@ func TestRetry_logger(t *testing.T) {
...
+	runTimeWithoutMaxElapsedTime := sleepInHandler * time.Duration(maxRetries)
+
+	h := retry.Middleware(func(msg *message.Message) (messages []*message.Message, e error) {
+		time.Sleep(sleepInHandler)
+		return nil, errors.New("foo")
+	})
+
+	startTime := time.Now()
+	_, _ = h(message.NewMessage("1", nil))
+	timeElapsed := time.Now().Sub(startTime)
```
golangcibot: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

message/router/middleware/retry.go
```diff
@@ -28,25 +26,39 @@ type Retry struct {
...
+
+			if r.shouldRetry(retries, elapsedTime) {
```
m110: How about changing this to avoid long nesting?

```go
if !r.shouldRetry(...) {
    return
}
```

m110: Or maybe `break` then and return at the end of the function.

PR #94, opened 2019-06-01
Middleware documentation

maclav3:
automatically generate hugo docs from godoc


message/router/middleware/ignore_errors.go
```diff
@@ -20,7 +24,7 @@ func (i IgnoreErrors) Middleware(h message.HandlerFunc) message.HandlerFunc {
 	return func(msg *message.Message) ([]*message.Message, error) {
 		events, err := h(msg)
 		if err != nil {
-			if _, ok := i.ignoredErrors[err.Error()]; ok {
+			if _, ok := i.ignoredErrors[errors.Cause(err).Error()]; ok {
```
roblaszczak: can you add a test?

message/router/middleware/instant_ack.go
```diff
@@ -2,6 +2,7 @@ package middleware
 
 import "github.com/ThreeDotsLabs/watermill/message"
 
+// InstantAck makes the handler instantly acknowledge the incoming message, regardless of any errors.
```
roblaszczak: some information that it can be useful for increasing throughput, when order and at least once delivery is ok for us

message/router/middleware/retry.go
```diff
@@ -10,6 +10,8 @@ import (
 
 type OnRetryHook func(retryNum int, delay time.Duration)
 
+// Retry provides a middleware that retries the handler if errors are returned.
+// The retry behaviour is configurable, with exponential backoff and maximum elapsed time.
 type Retry struct {
 	MaxRetries int
```
roblaszczak: it would be nice to have docs also for params

message/router/middleware/ignore_errors.go
maclav3: faa2f17941a31df27f6ec0ca7a5f2a1b91b74203 there you go

message/router/middleware/retry.go
maclav3: you do that, you wrote the middleware xD
but more seriously, I'll try to come up with something

maclav3: I've updated the retry middleware to use cenkalti/backoff
4538ec1ea4a94403d70a976f5ab9b828091840f7

message/router/middleware/instant_ack.go
maclav3: 4c18a1ce2169c218bc64caa54c3d6875f248758c

PR #97, opened 2019-06-14
update sql pubsub

roblaszczak:


message/infrastructure/sql/subscriber.go
```diff
@@ -89,7 +101,13 @@ func NewSubscriber(db *sql.DB, config SubscriberConfig) (*Subscriber, error) {
 		return nil, errors.Wrap(err, "invalid config")
 	}
 
+	idBytes, idStr, err := newSubscriberID()
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

message/infrastructure/sql/offsets_adapter.go
```diff
@@ -0,0 +1,64 @@
...
+		VALUES (?, ?) ON DUPLICATE KEY UPDATE offset=VALUES(offset)`
+
+	return ackQuery, []interface{}{offset, consumerGroup}
+}
+
+func (s DefaultOffsetsAdapter) NextOffsetQuery(topic, consumerGroup string) (string, []interface{}) {
+	return `
+		SELECT COALESCE(MAX(` + s.MessagesOffsetsTable(topic) + `.offset), 0)
+		FROM ` + s.MessagesOffsetsTable(topic) + `
+		WHERE consumer_group=? FOR UPDATE`, []interface{}{consumerGroup}
```
roblaszczak: added `FOR UPDATE`, it should help to handle multiple subscribers (todo - test)

message/infrastructure/sql/publisher.go
```diff
@@ -19,6 +20,10 @@ type PublisherConfig struct {
 
 	// SchemaAdapter provides the schema-dependent queries and arguments for them, based on topic/message etc.
 	SchemaAdapter SchemaAdapter
+
+	// AutoInitializeSchema enables initialization of schema database during publish.
+	// Schema is initialized once per topic per publisher instance.
+	AutoInitializeSchema bool
```
roblaszczak: looked useful for example

message/infrastructure/sql/publisher.go
```diff
@@ -76,43 +84,61 @@ func NewPublisher(db db, config PublisherConfig) (*Publisher, error) {
...
+	if err != nil {
+		return errors.Wrap(err, "cannot create insert query")
+	}
+
+	p.config.Logger.Trace("Inserting message to SQL", watermill.LogFields{
+		"query":      insertQuery,
+		"query_args": sqlArgsToLog(insertArgs),
 	})
 
-	stmt, err := p.db.Prepare(insertQuery)
```
roblaszczak: I removed all statements, removing them made code faster (we don't need to call sql twice) and cleaner

message/infrastructure/sql/offsets_adapter.go
```diff
@@ -0,0 +1,64 @@
+package sql
+
+type OffsetsAdapter interface {
+	// AckQuery returns the SQL query that will mark a message as read for a given consumer group.
+	// Subscriber will not return those messages again for this consumer group.
+	AckMessageQuery(topic string, offset int, consumerGroup string) (string, []interface{})
```
roblaszczak: I changed approach to return query and args from one method, for me looks cleaner, what do you think?

message/infrastructure/sql/schema_adapter.go
```diff
@@ -0,0 +1,122 @@
...
+		"`created_at` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,",
+		"`payload` JSON DEFAULT NULL,",
+		"`metadata` JSON DEFAULT NULL",
+		");",
+	}, "\n")
+
+	return []string{createMessagesTable}
+}
+
+func (s DefaultSchema) InsertQuery(topic string, msgs message.Messages) (string, []interface{}, error) {
```
roblaszczak: I added support for multi insert with one query, it should be much faster

message/infrastructure/sql/schema.go
```diff
@@ -1,167 +1,40 @@
 package sql
 
 import (
-	"database/sql"
```
roblaszczak: stupid diff :/

message/infrastructure/sql/subscriber.go
```diff
@@ -64,13 +70,19 @@ func (c SubscriberConfig) validate() error {
...
+		return errors.New("offsets adapter is nil")
+	}
 
 	return nil
 }
 
 // Subscriber makes SELECT queries on the chosen table with the interval defined in the config.
 // The rows are unmarshaled into Watermill messages.
 type Subscriber struct {
+	consumerIdBytes  []byte
```
roblaszczak: this is innovation for the future: it can be used for making active master HA model (without changing contract)

message/infrastructure/sql/subscriber.go
```diff
@@ -200,46 +213,70 @@ func (s *Subscriber) query(
...
-	if acked {
-		var ackArgs []interface{}
-		ackArgs, err = s.config.SchemaAdapter.AckArgs(offset, s.config.ConsumerGroup)
+	consumedQuery, consumedArgs := s.config.OffsetsAdapter.ConsumedMessageQuery(
+		topic,
+		offset,
+		s.config.ConsumerGroup,
+		s.consumerIdBytes,
+	)
+	if consumedQuery != "" {
```
roblaszczak: this innovation may be useful for more performant offsets adapter

message/infrastructure/sql/subscriber.go
```diff
@@ -200,46 +213,70 @@ func (s *Subscriber) query(
...
 		}
 
-		logger.Trace(ackQuery, watermill.LogFields{
-			"args": fmt.Sprintf("%+v", ackArgs),
+		ra, err := result.RowsAffected()
+		if err != nil {
+			return errors.Wrap(err, "cannot get rows affected for consumed query")
+		}
+		if ra == 0 {
+			// TODO - TEST
```
roblaszczak: fyi: in progress

message/infrastructure/sql/offsets_adapter.go
maclav3: :+1:

message/infrastructure/sql/offsets_adapter.go
maclav3: It's great, I really like how they've become more concise.
The only problem I have is that the return types `(string, []interface{})` are not immediately evindent from the method name or the godoc.

`XQueryAndArgs` or `XQueryArgs` doesn't seem like a great idea. So maybe in the godoc we could add something like `"AckMessageQuery" returns the SQL query and arguments that will mark a message as read...`

A little bit clearer and doesn't hinder readability that much

maclav3: Also, the godoc starts with a different name than the actual method name

maclav3: Also, the return types could be named `(query string, args []interface{})`.

message/infrastructure/sql/publisher.go
maclav3: :+1:

message/infrastructure/sql/publisher.go
```diff
@@ -38,6 +43,7 @@ func (c *PublisherConfig) setDefaults() {
 // db is implemented both by *sql.DB and *sql.Tx
 type db interface {
 	Prepare(q string) (*sql.Stmt, error)
```
maclav3: If all statements are removed,, `Prepare` is not needed anymore in this interface?

message/infrastructure/sql/schema_adapter.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/ThreeDotsLabs/watermill/message"
+	"github.com/pkg/errors"
+)
+
+// SchemaAdapter produces the SQL queries and arguments appropriately for a specific schema and dialect
+// It also transforms sql.Rows into Watermill messages.
+type SchemaAdapter interface {
+	// InsertQuery returns the SQL query that will insert the Watermill message into the SQL storage.
+	InsertQuery(topic string, msgs message.Messages) (string, []interface{}, error)
```
maclav3: The comments under `OffsetsAdapter` apply here as well

message/infrastructure/sql/schema_adapter.go
maclav3: Each message adds three args to the slice. I don't think I'm worried about the golang slice limits, but maybe we could write a test to see how many messages the sql driver can handle and if something stupid and unexpected happens with some relatively sane number.

maclav3: If there's a panic for example, add some safeguards.

message/infrastructure/sql/subscriber.go
```diff
@@ -200,46 +214,70 @@ func (s *Subscriber) query(
...
 
-	row := selectStmt.QueryRowContext(ctx, selectArgs...)
-
-	var offset int
-	var msg *message.Message
-	offset, msg, err = s.config.SchemaAdapter.UnmarshalMessage(row)
+	offset, msg, err := s.config.SchemaAdapter.UnmarshalMessage(row)
 	if errors.Cause(err) == sql.ErrNoRows {
 		// wait until polling for the next message
+		logger.Debug("No more messages, waiting until next query", watermill.LogFields{
```
maclav3: this may get really annoying if there's little traffic and the interval is quite low.
how about trace maybe?

maclav3: i mean, it isn't classic "tracing" use of trace, but I guess it fits the use case?

message/infrastructure/sql/subscriber.go
```diff
@@ -200,46 +214,70 @@ func (s *Subscriber) query(
...
-		var ackArgs []interface{}
-		ackArgs, err = s.config.SchemaAdapter.AckArgs(offset, s.config.ConsumerGroup)
+	consumedQuery, consumedArgs := s.config.OffsetsAdapter.ConsumedMessageQuery(
+		topic,
+		offset,
+		s.config.ConsumerGroup,
+		s.consumerIdBytes,
+	)
+	if consumedQuery != "" {
+		logger.Trace("Executing consumed query", watermill.LogFields{
```
maclav3: "consumed query" isn't very explicit about what it is supposed to do.
it's in the trace level, so it's not a big problem, but maybe:
`Executing query to confirm message consumed` or sth similar?

message/infrastructure/sql/sql.go
```diff
@@ -0,0 +1,27 @@
...
+	for _, arg := range s {
+		strArgs = append(strArgs, fmt.Sprintf("%s", arg))
+	}
+
+	return strings.Join(strArgs, ",")
+}
+
+func isDeadlock(err error) bool {
+	// ugly, but should be universal for multiple sql implementations
+	if strings.Contains(strings.ToLower(err.Error()), "deadlock") {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

message/infrastructure/test_pubsub.go
```diff
@@ -141,6 +150,68 @@ func TestPubSubStressTest(
...
+func (s *subscriberMultiplier) Subscribe(ctx context.Context, topic string) (msgs <-chan *message.Message, err error) {
+	defer func() {
+		if err != nil {
+			if closeErr := s.Close(); closeErr != nil {
+				err = multierror.Append(err, closeErr)
+			}
+		}
+	}()
+
+	out := make(chan *message.Message, 0)
```
golangcibot: S1019: should use make(chan *message.Message) instead (from `gosimple`)

internal/subscriber/multiplier.go
```diff
@@ -0,0 +1,80 @@
...
+func (s *multiplier) Subscribe(ctx context.Context, topic string) (msgs <-chan *message.Message, err error) {
+	defer func() {
+		if err != nil {
+			if closeErr := s.Close(); closeErr != nil {
+				err = multierror.Append(err, closeErr)
+			}
+		}
+	}()
+
+	out := make(chan *message.Message, 0)
```
golangcibot: S1019: should use make(chan *message.Message) instead (from `gosimple`)

message/infrastructure/sql/schema_adapter.go
roblaszczak: it's not our domain (like addresses, validation you remember :P)

message/infrastructure/sql/subscriber.go
roblaszczak: it's by default 1s, I think it is good for debug

message/infrastructure/sql/offsets_adapter.go
roblaszczak: added info in doc about args, better now?

message/infrastructure/sql/publisher.go
roblaszczak: removed

message/infrastructure/sql/schema_adapter.go
roblaszczak: done

message/infrastructure/sql/subscriber.go
roblaszczak: done

message/infrastructure/sql/offsets_adapter.go
maclav3: :+1:

message/infrastructure/sql/publisher.go
maclav3: :+1:

message/infrastructure/sql/subscriber.go
maclav3: ok

PR #102, opened 2019-06-22
Tracing support

roblaszczak:


sagikazarmark:
Feel free too use my https://github.com/sagikazarmark/ocwatermill repo if there is anything useful there. I've just pushed a trace branch with all the code I've written for tracing.

Also, keep in mind that [OpenTelemetry](https://opentelemetry.io/) is coming.

K-Phoen:
For those looking for OpenTelemetry support, https://github.com/voi-oss/watermill-opentelemetry might be interesting :)

sagikazarmark:
Nice!

PR #106, opened 2019-08-04
Fix stress test add added more tests debug

roblaszczak:


message/infrastructure/test_pubsub.go
```diff
@@ -801,7 +893,7 @@ func TestReconnect(t *testing.T, pub message.Publisher, sub message.Subscriber,
 			publishMessage <- struct{}{}
 
 			if _, shouldRestart := restartAfterMessages[i]; shouldRestart {
-				go restartServer(t, features)
+				go restartServer(t, tCtx.Features)
```
golangcibot: SA2002: the goroutine calls T.Fatal, which must be called in the same goroutine as the test (from `staticcheck`)

message/infrastructure/test_pubsub.go
```diff
@@ -27,22 +29,66 @@ var defaultTimeout = time.Second * 15
...
-		}
+func TestPubSub(
+	t *testing.T,
+	features Features,
+	pubSubConstructor PubSubConstructor,
+	consumerGroupPubSubConstructor ConsumerGroupPubSubConstructor,
+) {
+	testFuncs := []struct {
+		Func        func(t *testing.T, tCtx TestContext, pubSubConstructor PubSubConstructor)
+		NotParallel bool
```
m110: Consider `Sequentional` instead of `NotParallel`.

PR #108, opened 2019-08-05
Better example verification

jedruniu:
What changed:

1. Validation moved from bash script to go program, script didn't worked on MacOS anyway
2. Config format changed to .yml files
3. Teardown command added, because if check is successful we don't want to wait for whole docker-compose to finish it's work


dev/validate-examples/main.go
```diff
@@ -0,0 +1,118 @@
...
+	validationCmd := exec.Command(cmdAndArgs[0], cmdAndArgs[1:]...)
+	validationCmd.Dir = filepath.Dir(path)
+	defer func() {
+		if config.TeardownCmd == "" {
+			return
+		}
+		cmdAndArgs := strings.Fields(config.TeardownCmd)
+		teardownCmd := exec.Command(cmdAndArgs[0], cmdAndArgs[1:]...)
+		teardownCmd.Dir = filepath.Dir(path)
+		teardownCmd.Run()
```
golangcibot: Error return value of `teardownCmd.Run` is not checked (from `errcheck`)

dev/validate-examples/main.go
```diff
@@ -0,0 +1,118 @@
...
+	err = yaml.Unmarshal(file, c)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func main() {
+	err := filepath.Walk(".", func(exampleConfig string, f os.FileInfo, err error) error {
+		matches, err := filepath.Match(".validate_example*.yml", f.Name())
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

dev/validate-examples/main.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	err = yaml.Unmarshal(file, c)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func main() {
+	err := filepath.Walk(".", func(exampleConfig string, f os.FileInfo, err error) error {
```
golangcibot: SA4009: argument err is overwritten before first use (from `staticcheck`)

dev/validate-examples/main.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	err = yaml.Unmarshal(file, c)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func main() {
+	walkErr := filepath.Walk(".", func(exampleConfig string, f os.FileInfo, err error) error {
```
golangcibot: SA4009: argument err is overwritten before first use (from `staticcheck`)

dev/validate-examples/main.go
```diff
@@ -0,0 +1,117 @@
...
+	"regexp"
+	"strings"
+	"time"
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+type Config struct {
+	ValidationCmd  string `yaml:"validation_cmd"`
+	TeardownCmd    string `yaml:"teardown_cmd"`
```
roblaszczak: instead of TeardownCmd you can use `exec.Cmd.Process` to just kill the process :)

dev/validate-examples/main.go
```diff
@@ -0,0 +1,117 @@
+package main
```
roblaszczak: it's probably a good idea to add separated go.mod here, to not affect entire package

dev/validate-examples/main.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func main() {
+	walkErr := filepath.Walk(".", func(exampleConfig string, f os.FileInfo, err error) error {
+		matches, _ := filepath.Match(".validate_example*.yml", f.Name())
+		if matches {
```
roblaszczak: it's always a good idea to keep a happy path at left ;) 
good article about that: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

dev/validate-examples/main.go
jedruniu: I would argue that `docker-compose down` is slightly better because:

- it removes stopped containers
- it removed network created for purpose of  corresponding `docker-compose up`

I mean - sure why not - but the way it is done right now leaves workspace cleaner. I am not sure we care that much but it is something to consider.

dev/validate-examples/main.go
jedruniu: Totally agree, changing

dev/validate-examples/main.go
roblaszczak: ok, true :)

PR #113, opened 2019-08-18
Update documentation and examples

m110:
* Added SQL documentation
* Getting started refreshed
* Added two SQL PubSub examples
* Fixed middleware docs generation (incorrect regexp)
* Minor docs updates and cleanups
* Removed global docker-comose.yml, as tests should not need it anymore (should speed up CI)


_examples/sql/event-log/README.md
```diff
@@ -0,0 +1,51 @@
+# Event log
+
+This example shows how to use the SQL Publisher from [SQL Pub/Sub](https://github.com/ThreeDotsLabs/watermill-sql).
+
+## Background
+
+Some PubSubs (e.g. Kafka) come with support for storing processed messages, possibly even with no expiration date.
+This can be useful for audit purposes or to reply selected messages again in the future. But what if you'd like to use
+a PubSub that offers no storage and an event log is needed?
+
```
roblaszczak: maybe also info that it's useful when you don't have any pubsub infrastructue, but just SQL?

_examples/sql/event-log/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Solution
+
+Plugging a pair of a publisher and a subscriber into Watermill's `Router` can work as a proxy from one PubSub to another.
+To ensure that events are written to a persistent storage, you can use the SQL publisher.
+
+Google Cloud Pub/Sub subscriber consumes events from a topic and inserts them into a MySQL table. While this particular
```
roblaszczak: so maybe it would be better in other direction, from sql to pubsub?

roblaszczak: ah ok, I see that it is another example with an approach like that :+1:

_examples/sql/transactional-events/main.go
```diff
@@ -0,0 +1,153 @@
...
+
+	return pub
+}
+
+type event struct {
+	Name       string `json:"name"`
+	OccurredAt string `json:"occurred_at"`
+}
+
+func simulateEvents(db *stdSQL.DB) {
```
roblaszczak: it would be nice to do it in the transaction (this is not obvious, but you need to create subscriber per transaction), IMO it should be also linked in the docs

docs/content/docs/messages-router.md
```diff
@@ -75,32 +71,33 @@ To run the Router, you need to call `Run()`.
...
 
 {{% render-md %}}
 {{% load-snippet-partial file="content/src-link/message/router.go" first_line_contains="// Running" last_line_contains="func (r *Router) Running()" padding_after="0" %}}
 {{% /render-md %}}
 
 ### Execution model
 
-Some *Consumers* may support only a single stream of messages - this means that until a `msg.Ack()` is sent, you will not receive any more messages.
+*Subscribers* can consume either one message at a time or multiple messages in parallel. Single stream of messages is the simplest approach
+and it means that until a `msg.Ack()` is called, the subscriber will not receive any more messages.
```
roblaszczak: > and it means that until a `msg.Ack()` is called, the subscriber will not receive any more messages.
it's not so simple, unfortunately :( 

for example, if you are consuming multiple partitions in Kafka, you will receive multiple messages in parallel, event without ack - so, in general, it depends on pub/sub implementation

roblaszczak: ah ok, it's below, maybe we can rephrase it a bit? because until you don't go below its not clear

roblaszczak: I was also thinking about a simple schema of how it's working, but maybe its overkill?

docs/content/pubsubs/sql.md
```diff
@@ -0,0 +1,76 @@
...
+{{% load-snippet-partial file="src-link/watermill-sql/pkg/sql/publisher.go" first_line_contains="// Publish " last_line_contains="func (p *Publisher) Publish" %}}
+{{% /render-md %}}
+
+#### Subscribing
+
+{{% render-md %}}
+{{% load-snippet-partial file="src-link/watermill-sql/pkg/sql/subscriber.go" first_line_contains="func (s *Subscriber) Subscribe" last_line_contains="func (s *Subscriber) Subscribe" %}}
+{{% /render-md %}}
+
+#### Schema
```
roblaszczak: I think that it should be more explicitly mentioned that if you need, you can easily implement your own schema or offset schema (IMO it need to be done pretty often, to have a bit different schema), and that already implemented one was tested with MySQL.

It's also probably good to mention that by using composition it can be done easier.

_examples/sql/event-log/README.md
m110: I'll add that one to the general SQL docs, as this example is focused on a particular problem.

We could also add a third example using just the SQL pub/sub, but there's already something like this in the getting started section, so I feel that would be overkill.

docs/content/docs/messages-router.md
m110: See 6b1b00e6377f7292d59ad1ab6828b5ef770f3db0

A schema could be nice, although I don't have an idea at the moment how to present this clearly. But the refined docs should be now pretty easy to understand.

docs/content/pubsubs/sql.md
m110: See 9a80672071fda8e26574a88f0a00bd97388acc3e

_examples/sql/transactional-events/main.go
m110: See 9a80672071fda8e26574a88f0a00bd97388acc3e

roblaszczak: IMO it would be also nice to have a comment mentioning that creating publisher per transaction is ok (both in godoc and in examples)

docs/content/docs/messages-router.md
```diff
@@ -75,32 +71,36 @@ To run the Router, you need to call `Run()`.
...
-However, some *Consumers* support multiple streams. For example, by subscribing to multiple partitions in parallel,
-multiple messages will be sent, even previous that were not Acked (e.g. the Kafka Consumer works like this).
-The router can handle this case and spawn multiple `HandlerFunc` in parallel.
+* **Single stream of messages** is the simplest approach and it means that until a `msg.Ack()` is called, the subscriber
+  will not receive any new messages.
+* **Multiple message streams** are supported only by some subscribers. By subscribing to multiple topic partitions at once,
+  several messages can be consumed in parallel, even previous messages that were not acked (for example, the Kafka subscriber
+  works like this). Router handles this model by running concurrent `HandlerFunc`s, one for each partition.
+  
+See the chosen Pub/Sub documentation for supported execution models.
```
roblaszczak: just simple idea: how easy it is to find it in pub/sub's docs? but it's probably for another issue...

_examples/sql/transactional-events/main.go
roblaszczak: ok, I see that it's already there :+1:

docs/content/docs/messages-router.md
roblaszczak: So maybe just an example how to add (for example different UUID format BIN(16) or something like that)

_examples/sql/transactional-events/main.go
```diff
@@ -0,0 +1,171 @@
...
+		if err != nil {
+			panic(err)
+		}
+
+		time.Sleep(time.Second)
+	}
+}
+
+func publishEvent(tx *stdSQL.Tx) error {
+	pub, err := sql.NewPublisher(tx, sql.PublisherConfig{
```
maclav3: we could use a comment like 
`// we create a new publisher for each event, in order for the event publication to be transactional`, or something better worded?

docs/content/docs/getting-started.md
```diff
@@ -125,21 +173,63 @@ Running in Docker
...
+{{% load-snippet-partial file="content/docs/getting-started/sql/main.go" first_line_contains="package main" last_line_contains="process(messages)" %}}
+{{% load-snippet-partial file="content/docs/getting-started/sql/main.go" first_line_contains="func process" %}}
+{{% /tabs-tab %}}
+
 {{% /tabs %}}
 
-### Publishing messages
+### Creating Messages
+
+Watermill doesn't enforce any message format. `NewMessage` expectes a slice of bytes as the payload. You can use
```
maclav3: expects*

docs/content/docs/getting-started.md
```diff
@@ -125,21 +173,63 @@ Running in Docker
...
+The message UUID is optional, but recommended, as it helps with debugging.
 
-{{% tabs id="publishing" tabs="go-channel,kafka,nats-streaming,gcloud,amqp" labels="Go Channel,Kafka,NATS Streaming,Google Cloud Pub/Sub,RabbitMQ (AMQP)" %}}
+```go
+msg := message.NewMessage(watermill.NewUUID(), []byte("Hello, world!"))
+```
+
+### Publishing Messages
+
+`Publish` expects a topic and a `Message` to be published.
```
maclav3: `msg` is variadic, so rather:
```
	`Publish` expects a topic and one or more `Message`s to be published.
```

maclav3: btw, if it's variadic, then 0 messages may be passed.
I don't exactly know how the implementations handle an empty slice of messages, but I guess most just iterate over `[]` and `return nil`?

docs/content/docs/getting-started.md
m110: See 131b47695b6334095112ccef6d08d3f9e8ab741f

_examples/sql/transactional-events/main.go
m110: Added a comment in 131b47695b6334095112ccef6d08d3f9e8ab741f

docs/content/docs/getting-started.md
m110: Good point, added "one or more"  in 131b47695b6334095112ccef6d08d3f9e8ab741f

Most implementations will just iterate over the messages and don't do anything if the slice is empty. But I think it's not worth mentioning in the getting started, to keep it easy to grasp for new users. It's also an implementation detail that can vary between PubSubs, while the variadic `Message` is part of Watermill's interface.

docs/content/docs/messages-router.md
m110: Added general instruction in fe30835c840bd8be368e599cdb8bf7dd3e3c20aa
The example extended code is shown within the included snippet.

I think sometime in the future we can add yet another SQL example, showing how to prepare full schema and offset adapters.

_examples/sql/transactional-events/main.go
maclav3: perfect <3

_examples/sql/event-log/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+After few seconds, some events should be saved in the table:
+
+```
+docker-compose exec mysql mysql -e 'select * from watermill.watermill_events;'
++--------+--------------------------------------+---------------------+---------------------+----------+
+| offset | uuid                                 | created_at          | payload             | metadata |
++--------+--------------------------------------+---------------------+---------------------+----------+
+|      1 | 2faf6a14-f52a-4d6c-a4be-7355db428be1 | 2019-08-17 12:23:35 | {"num": 1566044614} | {}       |
```
maclav3: the payload in this example has `Name` and `OccuredAt`, while this table has `"num"`
Did I miss something or is it a bad copypaste?

m110: I've updated the code but not the README. However now the table takes too much space to show it nicely, so I've just cut the payload: be799a0ab8e0d60b3cb5177b8fe3f03c70316900

maclav3: :+1:

PR #114, opened 2019-08-24
Run example with context

gabeduke:


m110:
Thanks @gabeduke! The examples are slightly outdated right now, but will be all fixed up before the 1.0 release.

PR #117, opened 2019-08-28
Print stacktrace on panic recover.

andreyromancev:
Hey there! We had some problems with finding panic reason in our code. Added stack trace to error log.

### DIFF
--- a/message/router.go
+++ b/message/router.go
@@ -3,6 +3,7 @@ package message
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


m110:
Hey @andreyromancev. Thanks for contributing! Was the panic related with code running in the handler? You can use the Recoverer middleware as well to catch errors like this: https://github.com/ThreeDotsLabs/watermill/blob/master/message/router/middleware/recoverer.go

PR #119, opened 2019-08-29
Fix getting-started logs

jedruniu:
I beleive there is a mistake in the doc - here is the fix.


m110:
Thanks @jedruniu. We've got a bigger docs update in progress, see #113. This paragraph has been cleaned up there.

PR #120, opened 2019-08-31
Update simple-app example

m110:
* Add "basic" examples category and move `your-first-app` there (no changes in code).
* Rename `simple-app` to `realtime-feed` and move it to the basic category.
* Add some background in the README.
* Usual cosmetic style updates, grammar fixes and comments.
* Added "exercises" section to the README, new idea that I think might be also added in other examples. They are usually kept simple to avoid clutter, but can be easily extended.


_examples/basic/2-realtime-feed/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+To run this example you will need Docker and docker-compose installed. See the [installation guide](https://docs.docker.com/compose/install/).
+
+## Running
+
+```bash
+docker-compose up
+```
+
+You should see the live feed of produces posts on the standard output.
```
maclav3: typo
```suggestion
You should see the live feed of produced posts on the standard output.
```
or drop the 'produced' altogether, it's not essential

_examples/basic/2-realtime-feed/producer/main.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+			if messagesToPublish == 0 {
+				allMessagesPublished <- struct{}{}
+			}
+		}
+	}()
+
+	for num := 0; num < workers; num++ {
+		go func() {
+			for messagesToPublish > 0 {
```
maclav3: Isn't this going to race with the write at 38? it's not synced with the other goroutine.

m110: Nice catch. 👍 Reworked to distribute work by channels in 3330bd4d8c107b3fdcbd5bc4f25369adedd17f48

maclav3: This could perhaps be made even clearer if `chan message.Message` were used instead of the struct channels?
I'll try to propose my version in a separate PR

maclav3: https://github.com/ThreeDotsLabs/watermill/pull/121/files

maclav3: I made some manual testing and it seems that all the messages are published. The channel is buffered to the number of workers, so `close` shouldn't be hit before all messages are consumed.

I'm not 100% sure that we are losing no message here, but I don't know if it's crucial in an example like this

_examples/basic/2-realtime-feed/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+You should see the live feed of posts on the standard output.
+
+## Exercises
+
+1. Peek into the posts counter published on `posts_count` topic.
+
+```
+docker-compose exec kafka kafka-console-consumer --bootstrap-server localhost:9092 --topic posts_count
```
roblaszczak: In theory, we have mill ;) but it's hard to run it with the one-liner, maybe for the future ;)

_examples/basic/2-realtime-feed/consumer/main.go
```diff
@@ -22,72 +22,72 @@ var (
 )
 
 func main() {
-	logger := watermill.NewStdLogger(true, true)
+	logger := watermill.NewStdLogger(false, false)
+	logger.Info("Starting the consumer", watermill.LogFields{})
```
roblaszczak:
```suggestion
	logger.Info("Starting the consumer", watermill.LogFields{})
```
```suggestion
	logger.Info("Starting the consumer", nil)
```

_examples/basic/2-realtime-feed/consumer/main.go
```diff
@@ -22,72 +22,72 @@ var (
...
-		// if retries limit was exceeded, message is sent to poison queue (poison_queue topic)
+		// Retry middleware retries message processing if an error occurred in the handler
 		retryMiddleware.Middleware,
 
-		// recovered recovers panic from handlers
+		// If the retries limit is exceeded, the message is sent to the poison queue (poison_queue topic)
+		poisonQueue,
+
+		// Recoverer middleware recovers panic from handlers
 		middleware.Recoverer,
```
roblaszczak: basically, it should be on top in case of above middlewares failure

_examples/basic/2-realtime-feed/producer/main.go
```diff
@@ -0,0 +1,106 @@
...
+	publisher, err := kafka.NewPublisher(brokers, kafka.DefaultMarshaler{}, nil, logger)
+	if err != nil {
+		panic(err)
+	}
+	defer publisher.Close()
+
+	messagePublished := make(chan struct{})
+	allMessagesPublished := make(chan struct{})
+
+	go func() {
```
roblaszczak: after some thinking: maybe we should remove logic to produce X messages, and instead produce X messages per second, forever?

roblaszczak: it may simplify it a bit

_examples/basic/2-realtime-feed/producer/main.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	for i := 0; i < messagesToPublish; i++ {
+		workerMessages <- struct{}{}
+	}
+
+	// Waiting for all messages to be published
+	<-allMessagesPublished
+
+	logger.Info("All messages published", watermill.LogFields{})
```
roblaszczak:
```suggestion
	logger.Info("All messages published", nil)
```

_examples/basic/2-realtime-feed/producer/main.go
roblaszczak: also, proposition from #121 is nice :+1:

_examples/basic/2-realtime-feed/README.md
m110: Nice idea, I'll consider adding a note in some way. I don't think we have mill covered anywhere in README or in the docs.

_examples/basic/2-realtime-feed/producer/main.go
m110: Yeah that makes sense, as now the producer publishes all of them quickly and then gets restarted by docker-compose. 👍

_examples/basic/2-realtime-feed/consumer/main.go
m110: Ah right, I've changed this because the comments were hard to read that way. I'll update it somehow.

m110: Ah no, I changed the poison queue and retry order. But I guess this is wrong as well.

m110: See 3b1f8a17e4013c7a05e5224fa4653a845de7c7fe

_examples/basic/2-realtime-feed/producer/main.go
m110: @maclav3 can you handle this one in #121?

_examples/basic/2-realtime-feed/README.md
m110: So we could add another service to the compose, but it gets kinda ugly running it inside compose, with dependencies and mounting volumes by relative path. It works, but I feel like using kafka-console-consumer can be a bit easier to understand by newcomers.

Instead, I think we could add a separate docs page for mill. What do you think?

_examples/basic/2-realtime-feed/producer/main.go
maclav3: done, merged

_examples/basic/2-realtime-feed/producer/main.go
maclav3: #121 merged.

_examples/basic/2-realtime-feed/README.md
roblaszczak: but maybe just mill without docker, and link to sub-directory (to have info how to install etc)

_examples/basic/2-realtime-feed/consumer/main.go
```diff
@@ -22,72 +22,73 @@ var (
...
 
 	poisonQueue, err := middleware.PoisonQueue(pub, "poison_queue")
 	if err != nil {
 		panic(err)
 	}
 
 	r.AddMiddleware(
-		// limiting processed messages to 10 per second
-		middleware.NewThrottle(100, time.Second).Middleware,
+		// Recoverer middleware recovers panic from handlers
```
roblaszczak:
```suggestion
		// Recoverer middleware recovers panic from handlers and middlewares
```

_examples/basic/2-realtime-feed/producer/main.go
roblaszczak: VERY NICE, I LIKE

_examples/basic/2-realtime-feed/README.md
roblaszczak: but maybe in another task :P added to todolist

PR #121, opened 2019-09-01
 Use chan *message.Message to simplify producing messages

maclav3:


_examples/basic/2-realtime-feed/producer/main.go
```diff
@@ -33,42 +34,14 @@ func main() {
...
-	messagePublished := make(chan struct{})
-	allMessagesPublished := make(chan struct{})
-
-	go func() {
-		publishedMessages := 0
-		for range messagePublished {
-			publishedMessages++
-
-			if publishedMessages%1000 == 0 {
-				logger.Info("messages left", watermill.LogFields{"count": messagesToPublish - publishedMessages})
```
maclav3: I accidentally the `messages left` counter.
It may be restored if needed

_examples/basic/2-realtime-feed/producer/main.go
```diff
@@ -33,42 +34,14 @@ func main() {
...
-			if publishedMessages >= messagesToPublish {
-				allMessagesPublished <- struct{}{}
-				break
-			}
-		}
-	}()
-
-	workerMessages := make(chan struct{})
+	messageStream := make(chan *message.Message, workers)
+	workersGroup := &sync.WaitGroup{}
```
m110: I think you can just do `wg.Add(workers)`, as using `Add` and `Wait` concurrently is considered a race.

_examples/basic/2-realtime-feed/producer/main.go
m110: @roblaszczak proposed a new approach here: https://github.com/ThreeDotsLabs/watermill/pull/120/files/3330bd4d8c107b3fdcbd5bc4f25369adedd17f48#diff-4a06235922e53c9d4510ba3e7a3257fd

So maybe you could change the logic a bit, so there are X messages produced per second? The counter can be then dropped (or another log added).

maclav3: 637a8b3c7b5510560fee8b42ecd710f5f06a8d54

_examples/basic/2-realtime-feed/producer/main.go
maclav3: 637a8b3c7b5510560fee8b42ecd710f5f06a8d54

PR #135, opened 2019-09-16
Fix handleMessage printing 'Message acked' for nacked messages in deferred func

maclav3:


message/router_test.go
```diff
@@ -164,6 +165,144 @@ func TestRouter_functional_nack(t *testing.T) {
...
+			topic := "topic" + tc.Name
+			router.AddHandler(
+				"handler"+tc.Name,
+				topic,
+				subscriber,
+				topic,
+				publisher,
+				handlerFunc,
+			)
+			go router.Run(context.Background())
```
golangcibot: Error return value of `router.Run` is not checked (from `errcheck`)

PR #138, opened 2019-09-19
Fix Link

garciaolais:
Fix 
- Your first app
- Simple application with publisher and subscriber


m110:
Thank you @garciaolais. We're shuffling examples around in scope of #137, but they will be all up-to-date with the release of 1.0.0. :)

PR #139, opened 2019-09-21
Stylistic improvements

maclav3:


docs/content/docs/getting-started.md
```diff
@@ -262,16 +262,19 @@ if err != nil {
...
-The flow of our application looks like this:
+### Example application of *Messages Router*
+The flow of the example application looks like this:
 
-1. A message is produced on topic `example.topic_1` every second.
-2. `struct_handler` handler listens on `example.topic_1`. When a message is received, the UUID is printed and a new message is produced on `example.topic_2`.
-3. `print_events_topic_1` handler listens on `example.topic_1` and prints message UUID, payload and metadata. Correlation ID should be the same as in the message on `example.topic_1`.
-4. `print_events_topic_2` handler listens on `example.topic_2` and prints message UUID, payload and metadata. Correlation ID should be the same as in the message on `example.topic_2`.
+1. A message is produced on topic `incoming_messages_topic` every second.
+2. `struct_handler` handler listens on `incoming_messages_topic`. When a message is received, the UUID is printed and a new message is produced on `outgoing_messages_topic`.
```
m110: We've got two nicely named handlers now, so maybe we could rename this one as well?

maclav3: Thought about it.
The problem is that being a struct method (as opposed to a free-floating function) is quite a big part of this handler's identity, so I was reluctant to change the name.
I think if I named this one `incoming_messages_handler` and outlined the difference between func and method the subsection handlers well enough, this could work?

m110: Yeah, I think this is a good idea. Maybe also something along `routing_handler`, `first_handler` or `main_handler` might work, but I leave it up to you.

PR #145, opened 2019-09-25
Add thousands separator to benchmark results in README.md

danielparks:
The separator just makes it a little easier to understand quickly.


m110:
Thanks @danielparks!

Issue #133, opened 2019-09-16
Middleware per handler


roblaszczak:
Currently, you are not able to use middleware per handler in "clean way" (you can still manually wrap the handler function, or have separated router).

It would be nice to be able to add middleware per handler.

As a good reference point, you can use chi router: https://github.com/go-chi/chi

0michalsokolowski0:
I want to work on it

0michalsokolowski0:
PR: https://github.com/ThreeDotsLabs/watermill/pull/149

bkielbasa:
shouldn't the task be closed? :)

roblaszczak:
Yeah, done ;-)

PR #149, opened 2019-09-30
Middleware per handler #133

0michalsokolowski0:
- Implemented Middleware per handler
- Added tests

### DIFF
--- a/message/router.go
+++ b/message/router.go
@@ -99,10 +99,22 @@ func NewRouter(config RouterConfig, logger watermill.LoggerAdapter) (*Router, er
 	}, nil
 }
 
+type Middleware struct {
+	Handler     HandlerMiddleware
+	HandlerName string
+}
+
+func NewMiddleware(handler HandlerMiddleware, handlerName string) Middleware {
+	return Middleware{
+		Handler:     handler,
+		HandlerName: handlerName,
+	}
+}
+
 type Router struct {
 	config RouterConfig
 
-	middlewares []HandlerMiddleware
+	middlewares []Middleware
 
 	plugins []RouterPlugin
 
@@ -135,7 +147,10 @@ func (r *Router) Logger() watermill.LoggerAdapter {
 func (r *Router) AddMiddleware(m ...HandlerMiddleware) {
 	r.logger.Debug("Adding middlewares", watermill.LogFields{"count": fmt.Sprintf("%d", len(m))})
 
-	r.middlewares = append(r.middlewares, m...)
+	for _, handlerMiddleware := range m {
+		middleware := NewMiddleware(handlerMiddleware, "")
+		r.middlewares = append(r.middlewares, middleware)
+	}
 }
 
 func (r *Router) AddPlugin(p ...RouterPlugin) {
@@ -166,6 +181,14 @@ type DuplicateHandlerNameError struct {
 
 func (d DuplicateHandlerNameError) Error() string {
 	return fmt.Sprintf("handler with name %s already exists", d.HandlerName)
+}
+
+type HandlerNotExistError struct {
+	HandlerName string
+}
+
+func (d HandlerNotExistError) Error() string {
+	return fmt.Sprintf("handler with name %s not exists", d.HandlerName)
 }
 
 // AddHandler adds a new handler.
@@ -213,6 +236,27 @@ func (r *Router) AddHandler(
 		runningHandlersWg: r.runningHandlersWg,
 		messagesCh:        nil,
 		closeCh:           r.closeCh,
+	}
+}
+
+// AddMiddlewareToHandler adds a new middleware to specified handler in the router.
+//
+// The order of middlewares matters. Middleware added at the beginning is executed first.
+func (r *Router) AddMiddlewareToHandler(handlerName string, m ...HandlerMiddleware) {
+	r.logger.Debug("Adding middleware to handler", watermill.LogFields{
+		"count":       fmt.Sprintf("%d", len(m)),
+		"handlerName": handlerName,
+	})
+
+	_, ok := r.handlers[handlerName]
+
+	if !ok {
+		panic(HandlerNotExistError{handlerName})
+	}
+
+	for _, handlerMiddleware := range m {
+		middleware := NewMiddleware(handlerMiddleware, handlerName)
+		r.middlewares = append(r.middlewares, middleware)
 	}
 }
 
@@ -399,17 +443,20 @@ type handler struct {
 	closeCh chan struct{}
 }
 
-func (h *handler) run(middlewares []HandlerMiddleware) {
+func (h *handler) run(middlewares []Middleware) {
 	h.logger.Info("Starting handler", watermill.LogFields{
 		"subscriber_name": h.name,
 		"topic":           h.subscribeTopic,
 	})
 
 	middlewareHandler := h.handlerFunc
-
-	// first added middlewares should be executed first (so should be at the top of call stack)
-	for i := len(middlewares) - 1; i >= 0; i-- {
-		middlewareHandler = middlewares[i](middlewareHandler)
+	for i := 0; i < len(middlewares); i++ {
+		currentMiddleware := middlewares[i]
+		isGeneral := currentMiddleware.HandlerName == ""
+		isHandlerMiddleware := currentMiddleware.HandlerName == currentMiddleware.HandlerName
+		if isGeneral || isHandlerMiddleware {
+			middlewareHandler = currentMiddleware.Handler(middlewareHandler)
+		}
 	}
 
 	go h.handleClose()
--- a/message/router.go
+++ b/message/router.go
@@ -453,7 +453,7 @@ func (h *handler) run(middlewares []Middleware) {
 	for i := 0; i < len(middlewares); i++ {
 		currentMiddleware := middlewares[i]
 		isGeneral := currentMiddleware.HandlerName == ""
-		isHandlerMiddleware := currentMiddleware.HandlerName == currentMiddleware.HandlerName
+		isHandlerMiddleware := currentMiddleware.HandlerName == h.name
 		if isGeneral || isHandlerMiddleware {
 			middlewareHandler = currentMiddleware.Handler(middlewareHandler)
 		}
--- a/message/router.go
+++ b/message/router.go
@@ -183,14 +183,6 @@ func (d DuplicateHandlerNameError) Error() string {
 	return fmt.Sprintf("handler with name %s already exists", d.HandlerName)
 }
 
-type HandlerNotExistError struct {
-	HandlerName string
-}
-
-func (d HandlerNotExistError) Error() string {
-	return fmt.Sprintf("handler with name %s not exists", d.HandlerName)
-}
-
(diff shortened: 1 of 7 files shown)


message/router.go
```diff
@@ -399,17 +443,20 @@ type handler struct {
...
 
 	middlewareHandler := h.handlerFunc
-
-	// first added middlewares should be executed first (so should be at the top of call stack)
-	for i := len(middlewares) - 1; i >= 0; i-- {
-		middlewareHandler = middlewares[i](middlewareHandler)
+	for i := 0; i < len(middlewares); i++ {
+		currentMiddleware := middlewares[i]
+		isGeneral := currentMiddleware.HandlerName == ""
+		isHandlerMiddleware := currentMiddleware.HandlerName == currentMiddleware.HandlerName
```
golangcibot: SA4000: identical expressions on the left and right side of the '==' operator (from `staticcheck`)

message/router.go
```diff
@@ -216,6 +239,27 @@ func (r *Router) AddHandler(
 	}
 }
 
+// AddMiddlewareToHandler adds a new middleware to specified handler in the router.
+//
+// The order of middlewares matters. Middleware added at the beginning is executed first.
+func (r *Router) AddMiddlewareToHandler(handlerName string, m ...HandlerMiddleware) {
```
m110: Can you consider adding some high-level interface for this? I feel like basing this on handler name can be fragile and not that convenient.

Something similar to the go-chi way would be:

```go
router.Route(func (r Router) {
    r.AddMiddleware(middleware.Retry)
    r.AddHandler(myHandler)
})
```

But that can be quite hard to implement and maybe overcomplicated.

Another idea might be an interface similar to structured loggers:

```go
router.With(middleware.Retry).AddHandler(handler)

// Maybe even with variadic arguments
router.With(
    middleware.Recoverer,
    middleware.Retry,
).AddHandler(myHandler)
```

Any thoughts? :)

0michalsokolowski0: Thank you for the clarification.

In my opinion, the first approach might be a bit misleading since, in fact, there is no route (subpath) here. I personally prefer the second approach.

Could you tell me what is the expected flow of execution?
```
...
router.AddMiddleware(middlewareA)
router.AddHandler("handler_1",...)
router.AddMiddleware(middlewareB)
...
```
Currently (without middleware per handler) the flow looks like this:
`middlewareA => middlewareB => handler_1`

__What is the expected flow of the execution in such case?__
```
...
router.AddMiddleware(middlewareA)
router.With(
    middlewareC,
    middlewareD,
).AddHandler("handler_1",...)
router.AddMiddleware(middlewareB)
...
```

__Should it be:__
* `middlewareA => middlewareB => middlewareC => middlewareD => handler_1` or
* `middlewareA => middlewareC => middlewareD => middlewareB => handler_1` or
* `middlewareA => middlewareC => middlewareD => handler_1` (this one contradicts with the current solution, but it is similar to the approach used in go-chi where adding middleware after handler is forbidden https://github.com/go-chi/chi/blob/2171e2a8965db606ce84d224841987315c0eb515/mux.go#L93)

m110: That's interesting question and I'm not sure what's the most obvious for end users. I feel like the way chi handles it (with an error) seems the most idiomatic.

@roblaszczak @maclav3 any thoughts from you guys?

roblaszczak:
```
router.Route(func (r Router) {
    r.AddMiddleware(middleware.Retry)
    r.AddHandler(myHandler)
})
```

```router.With(middleware.Retry).AddHandler(handler)

// Maybe even with variadic arguments
router.With(
    middleware.Recoverer,
    middleware.Retry,
).AddHandler(myHandler)
```
this looks better, but IMO a bit wired (it can be confused with AddMiddleware)


and about order my guess is `middlewareA > MiddlewareC > MiddlewareD > Middleware B` (so we should probably have some global middleware counter?)


yup, it looks overcomplicated (in chi it make sense because you have sub-routers)

and different idea is, maybe we can return some builder with handler, so it may be

```
handler1 := router.AddHandler("handler_1", ...)
handler1.AddMiddleware(middlewareA)
```

message/router.go
```diff
@@ -399,17 +443,20 @@ type handler struct {
...
 		"subscriber_name": h.name,
 		"topic":           h.subscribeTopic,
 	})
 
 	middlewareHandler := h.handlerFunc
-
-	// first added middlewares should be executed first (so should be at the top of call stack)
-	for i := len(middlewares) - 1; i >= 0; i-- {
-		middlewareHandler = middlewares[i](middlewareHandler)
+	for i := 0; i < len(middlewares); i++ {
```
roblaszczak: so as we discussed - it should be done like before ;) (middlewares are applied in reversed order because the "outside" middleware is called first)

message/router.go
roblaszczak: so for the third option, I can imagine that AddHandler may return some kind of builder or `handler` object (with needs to be still connected to the router - to for example keep middlewares order) and may store internally middlewares added to handler

message/router.go
0michalsokolowski0: Yes, you are right I will remove the change.

message/router.go
0michalsokolowski0: @roblaszczak Thanks for the suggestions.

m110:
The new API can be added to the docs as well. :) Probably the best place is https://watermill.io/docs/messages-router/#middleware

message/router.go
```diff
@@ -99,10 +99,22 @@ func NewRouter(config RouterConfig, logger watermill.LoggerAdapter) (*Router, er
 	}, nil
 }
 
+type Middleware struct {
+	Handler     HandlerMiddleware
+	HandlerName string
+}
+
+func NewMiddleware(handler HandlerMiddleware, handlerName string) Middleware {
```
m110: This function probably doesn't have to be exported, as it's used only internally by the router.

m110: Actually, the `Middleware` struct can be private as well. It could be confusing to the end user why there's public `Middleware` and `HandlerMiddleware`.

message/router.go
```diff
@@ -135,7 +147,10 @@ func (r *Router) Logger() watermill.LoggerAdapter {
 func (r *Router) AddMiddleware(m ...HandlerMiddleware) {
 	r.logger.Debug("Adding middlewares", watermill.LogFields{"count": fmt.Sprintf("%d", len(m))})
 
-	r.middlewares = append(r.middlewares, m...)
+	for _, handlerMiddleware := range m {
+		middleware := NewMiddleware(handlerMiddleware, "")
```
m110: We don't enforce at the moment that the handler name is not empty, won't this be confusing if someone would create handler with empty name? Maybe it would be slightly cleaner to introduce some explicit boolean flag on the handler struct for this purpose.

message/router.go
```diff
@@ -185,7 +200,7 @@ func (r *Router) AddHandler(
 	publishTopic string,
 	publisher Publisher,
 	handlerFunc HandlerFunc,
-) {
+) *handler {
```
m110: `handler` is an unexported struct right now, so it might be better to not return it outside of the package (it's only for internal use right now). Instead, I would suggest to add new exported struct, that will hold only handler and router and will have a single public method `AddMiddleware`.

message/router.go
```diff
@@ -430,6 +456,21 @@ func (h *handler) run(middlewares []HandlerMiddleware) {
...
+// The order of middlewares matters. Middleware added at the beginning is executed first.
+func (h *handler) AddMiddleware(m ...HandlerMiddleware) {
+	h.logger.Debug("Adding middleware to handler", watermill.LogFields{
+		"count":       fmt.Sprintf("%d", len(m)),
+		"handlerName": h.name,
+	})
+
+	for _, handlerMiddleware := range m {
+		middleware := NewMiddleware(handlerMiddleware, h.name)
+		h.router.middlewares = append(h.router.middlewares, middleware)
```
m110: This looks like a violation of the Law of Demeter, it would be a bit more clean to introduce a method for this in router.

message/router.go
```diff
@@ -397,19 +423,24 @@ type handler struct {
 	messagesCh <-chan *Message
 
 	closeCh chan struct{}
+
+	router *Router
```
m110: Seems like this one is no longer needed.

message/router.go
```diff
@@ -430,6 +461,31 @@ func (h *handler) run(middlewares []HandlerMiddleware) {
 	h.logger.Debug("Router handler stopped", nil)
 }
 
+type Handler struct {
+	router  *Router
+	handler *handler
+}
+
+func NewHandler(router *Router, handler *handler) *Handler {
```
roblaszczak: it probably may be private ;)

roblaszczak: or even we can create it without constructor (it's used only in the package scope once, so it's fine)

message/router.go
```diff
@@ -99,10 +99,24 @@ func NewRouter(config RouterConfig, logger watermill.LoggerAdapter) (*Router, er
 	}, nil
 }
 
+type middleware struct {
+	Handler       HandlerMiddleware
+	HandlerName   string
+	IsRouterLevel bool
+}
+
+func newMiddleware(handler HandlerMiddleware, handlerName string, isRouterLevel bool) middleware {
```
roblaszczak: here also - we can live without constructor ;)

message/router.go
```diff
@@ -131,11 +145,18 @@ func (r *Router) Logger() watermill.LoggerAdapter {
...
-// The order of middlewares matters. Middleware added at the beginning is executed first.
+// The order of middlewares matters. middleware added at the beginning is executed first.
 func (r *Router) AddMiddleware(m ...HandlerMiddleware) {
 	r.logger.Debug("Adding middlewares", watermill.LogFields{"count": fmt.Sprintf("%d", len(m))})
 
-	r.middlewares = append(r.middlewares, m...)
+	r.addMiddleware("", true, m...)
+}
+
+func (r *Router) addMiddleware(handlerName string, isRouterLevel bool, m ...HandlerMiddleware) {
```
roblaszczak: in general IMO it's a good idea to avoid any bool in arguments and create separated functions here (and as I understand we need to pass handlerName only if it's handler level middleware, so it may also simplify it a bit

but in general, I like the design to keep everything in one slice :)

roblaszczak:
@0michalsokolowski0 I added a couple comments, but I like the current API, what do you think @m110?

It would be also nice to add some documentation of how to use it :)

m110:
Yeah I think the API is looking good now, I would maybe consider if we can come up with better names than `handler` and `Handler`. And docs would be nice :+1:

0michalsokolowski0:
@roblaszczak @m110 Please correct me if I am wrong but it seems to me that currently examples can only be updated after a release. So if I would like to extend [_examples/basic/3-router](https://github.com/ThreeDotsLabs/watermill/blob/master/_examples/basic/3-router/main.go) I can only do it after a release, since the example's dependencies point to specific version not the current implementation.

m110:
Yes, that's true. One way would be to add a draft PR, with go.mod pointing to this branch and after release just bump the version there.

Issue #131, opened 2019-09-16
"Duplicate chaos monkey" middleware


roblaszczak:
It would be nice, to have middleware that sends the message twice, to be sure that our handlers are idempotent.

0michalsokolowski0:
I would like to work on it

PR #156, opened 2019-10-28
Create middleware duplicating messages

0michalsokolowski0:
Resolving https://github.com/ThreeDotsLabs/watermill/issues/131

### DIFF
--- a/message/router/middleware/duplicator.go
+++ b/message/router/middleware/duplicator.go
@@ -1,15 +1,36 @@
 package middleware
 
 import (
+	"errors"
+
 	"github.com/ThreeDotsLabs/watermill/message"
+)
+
+var (
+	notIdempotentError = errors.New("handler is not idempotent")
 )
 
 // Duplicator sends message twice.
 func Duplicator(h message.HandlerFunc) message.HandlerFunc {
 	return func(msg *message.Message) ([]*message.Message, error) {
-		events, err := h(msg)
-		_, _ = h(msg)
+		firstProducedMessages, firstErr := h(msg)
+		secondProducedMessages, secondErr := h(msg)
+		allErrorsNil := firstErr == nil && secondErr == nil
+		allErrorsEqual := firstErr != nil && secondErr != nil && firstErr.Error() == secondErr.Error()
 
-		return events, err
+		if allErrorsNil || allErrorsEqual {
+			if len(firstProducedMessages) == len(secondProducedMessages) {
+				for index, msg := range firstProducedMessages {
+					messagesEqual := msg.Equals(secondProducedMessages[index])
+					if !messagesEqual {
+						return nil, notIdempotentError
+					}
+				}
+
+				return firstProducedMessages, nil
+			}
+		}
+
+		return nil, notIdempotentError
 	}
 }
--- a/message/router/middleware/duplicator.go
+++ b/message/router/middleware/duplicator.go
@@ -7,16 +7,16 @@ import (
 // Duplicator sends message twice.
 func Duplicator(h message.HandlerFunc) message.HandlerFunc {
 	return func(msg *message.Message) ([]*message.Message, error) {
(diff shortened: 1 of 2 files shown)


message/router/middleware/duplicator.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"github.com/ThreeDotsLabs/watermill/message"
+)
+
+// Duplicator sends message twice.
+func Duplicator(h message.HandlerFunc) message.HandlerFunc {
+	return func(msg *message.Message) ([]*message.Message, error) {
+		events, err := h(msg)
+		_, _ = h(msg)
```
m110: Handling of these errors is an interesting case. If we'd like to make sure that our system is idempotent, we should probably check if the second handler didn't return an error.

Also worth considering if other middlewares won't cause some weird issues.

message/router/middleware/duplicator.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Duplicator sends message twice.
+func Duplicator(h message.HandlerFunc) message.HandlerFunc {
+	return func(msg *message.Message) ([]*message.Message, error) {
+		firstProducedMessages, firstErr := h(msg)
+		secondProducedMessages, secondErr := h(msg)
+		allErrorsNil := firstErr == nil && secondErr == nil
+		allErrorsEqual := firstErr != nil && secondErr != nil && firstErr.Error() == secondErr.Error()
+
+		if allErrorsNil || allErrorsEqual {
```
roblaszczak: IMO we should keep simple responsibility here and just do duplication without checking returned messages and if erros are equal because messages may be a bit different, but still different after each call (for example: different correlation id's, uuids etc.)

We should also probably don't care about checking errors too much. IMO the best approach will be:
- if firstErr != nil, return (it anyway fails, so there is probably no point to try to fail it with duplication)
- if secondErr != nil - just return, something else should handle it

does it make sense? ;) @0michalsokolowski0

0michalsokolowski0: @roblaszczak You are right. Implementation changed in 105b2bbb1566a1cec843ffc14cf55c3b2654dc3a.

message/router/middleware/duplicator.go
```diff
@@ -0,0 +1,22 @@
+package middleware
+
+import (
+	"github.com/ThreeDotsLabs/watermill/message"
+)
+
+// Duplicator sends message twice.
```
roblaszczak:
```suggestion
// Duplicator is processing messages twice, to ensure that the endpoint is idempotent.
```

message/router/middleware/duplicator.go
```diff
@@ -0,0 +1,22 @@
...
+func Duplicator(h message.HandlerFunc) message.HandlerFunc {
+	return func(msg *message.Message) ([]*message.Message, error) {
+		producedMessages, firstErr := h(msg)
+		if firstErr != nil {
+			return nil, firstErr
+		}
+
+		_, secondErr := h(msg)
```
roblaszczak: I was thinking about it for a while, and probably we want to append it to `producedMessages`, eventually we can add config for this decorator (like in `Retry`) or separated decorator (like in `PoisonQueue` or `PoisonQueueWithFilter`).

message/router/middleware/duplicator.go
roblaszczak: much better now :medal_military:

message/router/middleware/duplicator.go
0michalsokolowski0: Fix c8baf8aad31ef594e153cc16f03f682864df26c4

0michalsokolowski0:
@roblaszczak @m110 Could you merge the PR? I don't have write access to the repository

m110:
@0michalsokolowski0 Merged, thanks for the contribution!

PR #166, opened 2019-11-28
Add FanOut proposal

m110:
During development of https://github.com/ThreeDotsLabs/watermill-http/pull/3 it occurred to me that we could abstract the "router" I created here, so this is a proposal of adding a `FanOut` component.

This is essentially a wrapper over GoChannel Pub/Sub that lets you send messages coming from upstream subscriber to any number of internal publishers. This might be useful if you need a lot of subscribers and don't want to create a separate subscription for each of them.

Keep in mind this is an early idea, so feedback is welcome. Docs and more tests are obviously missing.

I also considered another API for this, that will create its own router and will be started just like a router. Maybe both could be available.


message/router.go
```diff
@@ -34,6 +34,11 @@ type HandlerFunc func(msg *Message) ([]*Message, error)
 // NoPublishHandlerFunc is HandlerFunc alternative, which doesn't produce any messages.
 type NoPublishHandlerFunc func(msg *Message) error
 
+// ProxyHandler is a handler that passes the message unchanged from the subscriber to the publisher.
+var ProxyHandler HandlerFunc = func(msg *Message) ([]*Message, error) {
```
roblaszczak: maybe `PassthroughHandler` would better show what it is doing? probably @maclav3 can add something :P

pubsub/gochannel/fanout.go
```diff
@@ -0,0 +1,67 @@
...
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/ThreeDotsLabs/watermill"
+	"github.com/ThreeDotsLabs/watermill/message"
+)
+
+type FanOut struct {
+	pubSub *GoChannel
```
roblaszczak: maybe we can prefix `pubSub` and `router` with `internal`?

pubsub/gochannel/fanout.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/ThreeDotsLabs/watermill"
+	"github.com/ThreeDotsLabs/watermill/message"
+)
+
+type FanOut struct {
```
roblaszczak: it would be nice to have some godoc here before "undrafting" ;)

pubsub/gochannel/fanout.go
m110: 666e8e97d7b8637930ea7140a41fe12ed9b6bea4

message/router.go
m110: 666e8e97d7b8637930ea7140a41fe12ed9b6bea4

pubsub/gochannel/fanout.go
m110: Actually after thinking this through, I think only `pubSub` should be `internal`, as it is created in the constructor?

pubsub/gochannel/fanout.go
m110: Added.

message/router.go
```diff
@@ -34,6 +34,11 @@ type HandlerFunc func(msg *Message) ([]*Message, error)
 // NoPublishHandlerFunc is HandlerFunc alternative, which doesn't produce any messages.
 type NoPublishHandlerFunc func(msg *Message) error
 
+// PassthroughHandler is a handler that passes the message unchanged from the subscriber to the publisher.
+var PassthroughHandler HandlerFunc = func(msg *Message) ([]*Message, error) {
```
maclav3: You could use the alias `Messages` instead of `[]*Message`, as we've done elsewhere.
If you think it'll improve readability

message/router.go
maclav3: I was thinking about `NoopHandler` (not perfect, because it's not only noop, but also returns the incoming message) or `IdentityHandler` (probably too obscure and mathematical).
I thought about `TransparentHandler` or `TransmitHandler`, but it's also kinda meh

message/router.go
m110: `HandlerFunc` doesn't use it though, so maybe we should change them all at once?

message/router.go
m110: I like the proposed "passthrough", as the behavior is close to the logic gate with the same name. `NoopHandler` would look more like:

```go
func(msg *Message) ([]*Message, error) {
    return nil, nil
}
```

I even thought about adding it as well, but then again I can't think of any use for such handler.

message/router.go
maclav3: hmm I thought I saw some `HandlerFunc`s using `Messages` instead, maybe I was wrong

pubsub/gochannel/fanout.go
```diff
@@ -0,0 +1,99 @@
...
+	subscriber message.Subscriber
+
+	logger watermill.LoggerAdapter
+
+	subscribedTopics map[string]struct{}
+	subscribedLock   sync.Mutex
+}
+
+// NewFanOut creates new FanOut.
+// The passed router should not be running yet.
```
maclav3: In principle, we could check if the router is already running (`router.Running()`)?
Should we return `error` if this is the case?
Then cover it with the appropriate unit test ofc

m110: Oh right, that's a good idea. 👍

pubsub/gochannel/fanout.go
```diff
@@ -0,0 +1,99 @@
...
+func NewFanOut(
+	router *message.Router,
+	subscriber message.Subscriber,
+	logger watermill.LoggerAdapter,
+) (*FanOut, error) {
+	if router == nil {
```
roblaszczak: basically, is there any case when we need to pass router?

pubsub/gochannel/fanout.go
m110: See 6939c53598672004fbc4b4f704358d61574d432a

pubsub/gochannel/fanout.go
m110: Moved to internal router in a96ad451dc7bc6c5258ef8c684dbc6d7270af7a4

PR #170, opened 2019-12-17
Added more poison queue context

roblaszczak:


message/router/middleware/poison_test.go
```diff
@@ -106,6 +111,51 @@ func TestPoisonQueue_handler_failing(t *testing.T) {
...
+
+	pq, err := middleware.PoisonQueue(pubSub, "poison_queue")
+	require.NoError(t, err)
+	router.AddMiddleware(pq)
+
+	router.AddNoPublisherHandler("handler_name", "test", pubSub, func(msg *message.Message) error {
+		return errors.New("error")
+	})
+
+	go router.Run(context.Background())
```
golangcibot: Error return value of `router.Run` is not checked (from `errcheck`)

message/router/middleware/poison.go
```diff
@@ -11,6 +11,9 @@ var ErrInvalidPoisonQueueTopic = errors.New("invalid poison queue topic")
 
 // ReasonForPoisonedKey is the metadata key which marks the reason (error) why the message was deemed poisoned.
 var ReasonForPoisonedKey = "reason_poisoned"
+var PoisonedTopicKey = "topic_poisoned"
```
maclav3: those keys could be made consts to prevent surprises maybe

message/router/middleware/poison_test.go
```diff
@@ -106,6 +111,53 @@ func TestPoisonQueue_handler_failing(t *testing.T) {
...
+	err = pubSub.Publish("test", message.NewMessage("1", nil))
+	require.NoError(t, err)
+
+	msgs, err := pubSub.Subscribe(context.Background(), "poison_queue")
+	require.NoError(t, err)
+
+	messages, all := subscriber.BulkRead(msgs, 1, time.Second)
+	require.True(t, all, "no messages received")
+
+	assert.Equal(t, "handler_name", messages[0].Metadata[middleware.PoisonedHandlerKey])
```
maclav3: those strings could be declared before creating the handler, and referenced by the variable name in both places.
Apart from `gochannel.Gochannel`, that comes from some more complicated logic

message/router/middleware/poison.go
roblaszczak: changed

message/router/middleware/poison_test.go
roblaszczak: > those strings could be declared before creating the handler, and referenced by the variable name in both places.
IMO when it's used twice using variable doesn't help a lot and obfuscates it a bit

> Apart from gochannel.Gochannel, that comes from some more complicated logic
yes, but here I'm not interested with this logic probably, but just with the outcome

maclav3: fair enough

PR #178, opened 2020-02-19
Close all subscribers in TestConcurrentSubscribe

chrisseto:
I was running into some errors when building a new subscriber due to subscribers continuing to log after a test has finished.

This changes ensures that the created multi-subscriber will be closed at the end of `TestConcurrentSubscribe`


roblaszczak:
@chrisseto thanks for the contribution! :beers:

PR #183, opened 2020-04-10
fix context goChannel

xabi93:
Using goChannel I was trying to publish a message, previously I had added a value to the message context.

In the subscriber I tried to get that value but it wasn't. Checking the code I have seen that the msg context is replaced by other context `s.context`.

Shouldn't we use the same context?


jitendra-1217:
> Using goChannel I was trying to publish a message, previously I had added a value to the message context.

What kind of value(bit more details) were you putting in message's context?

roblaszczak:
Hello, it's valid behavior how it's working now ;) when you have "non-go channel" pubsub you are not able to send context via message brokers, because it's technically not possible, because of that ctx from the subscriber is passed. By design all pub/subs implementations should be compatible. This is why tests are failing ;)

If you need to pass some values I recommend using metadata.

PR #187, opened 2020-04-28
Add Forwarder component

czeslavo:
`Forwarder` is meant to automatically forward messages from one Pub/Sub to another (let's say _Input_ and _Output_), listening on a single _Input_ forwarder topic. Messages are put into an envelope when sent with a decorated publisher to _Input_ and then unwrapped by the forwarder and sent to _Output_ on a destination topic taken from the envelope. 

`Forwarder` uses `Router` underneath. Additionally it has `DecoratePublisher` helper method which decorates publisher meant to send messages to _Input PubSub_.


components/forwarder/forwarder.go
```diff
@@ -0,0 +1,88 @@
...
+	config.setDefaults()
+
+	router, err := message.NewRouter(message.RouterConfig{}, logger)
+	if err != nil {
+		return nil, err
+	}
+
+	f := &Forwarder{router, publisherOut, logger, config}
+
+	router.AddNoPublisherHandler(
```
maclav3, line 33: I was thinking 'why not `AddHandler`, but then realized that the topic is dynamic and needs to be unraveled from the envelope.
I guess a publisher decorator could take care of that, but probably readability might suffer

components/forwarder/forwarder.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// NewForwarder creates a forwarder which will subscribe to the topic provided in the provided config using the provided subscriber.
+// It will publish messages received on this subscription to the destination topic embedded in the enveloped message using the provided publisher.
+//
+// Provided subscriber and publisher can be from different Watermill Pub/Sub implementations, i.e. MySQL subscriber and Google Pub/Sub publisher.
+func NewForwarder(subscriberIn message.Subscriber, publisherOut message.Publisher, logger watermill.LoggerAdapter, config Config) (*Forwarder, error) {
+	config.setDefaults()
+
+	router, err := message.NewRouter(message.RouterConfig{}, logger)
```
maclav3, line 26: if we are using a router here, then we might want to use some middlewares as well
I'm thinking additional param `routerMiddlewares []Middleware` or additional config field.

We could also pass a router by constructor param, but would have to assert that it's not already running so that we can add the handler.

maclav3: But if we pass a router from outside, then `forwarder.Run()` and `forwarder.Close()` no longer make sense, because the router may be run or closed independently.
Probably better just to create a router in the constructor.

Otherwise, we'd have to turn the constructor into sth like `AddForwarderHandler(subscriberIn Subscriber, publisherOut Publisher, logger LoggerAdapter, config Config, router Router) error`

components/forwarder/forwarder.go
czeslavo: Yeah, I didn't think about using publisher decorator here, but I don't think it would give any additional value as the handler method is quite simple. I would opt for leaving it as it is for readability sake.

maclav3: 👍

components/forwarder/forwarder.go
```diff
@@ -0,0 +1,88 @@
...
+	return NewPublisherDecorator(publisher, f.config)
+}
+
+// AddSubscriberDecorators wraps the forwarder's Subscriber.
+func (f *Forwarder) AddSubscriberDecorators(dec ...message.SubscriberDecorator) {
+	f.router.AddSubscriberDecorators(dec...)
+}
+
+func (f *Forwarder) forwardMessage(msg *message.Message) error {
+	destTopic, unwrappedMsg, err := unwrapMessageFromEnvelope(msg)
```
maclav3, line 78: what if someone uses the undecorated input publisher?
I guess we would get an error here.
Will the faulty message NACK and retry in eternity?

I guess that's up to the middlewares of the router, see my other comment.

Alternatively, we could detect undecorated messages and silently ignore them, but maybe that's too opinionated.

Either way, sounds like a test case to me :)

components/forwarder/forwarder.go
czeslavo: According to our discussion last time, I have added middlewares to the forwarder's config. We will keep router as an internal detail and will not try to mimic its "API".

components/forwarder/forwarder.go
czeslavo: I have added a note regarding that and also covered this with a test case: 
a361c10

components/forwarder/forwarder.go
maclav3: 👍

components/forwarder/envelope.go
```diff
@@ -0,0 +1,74 @@
...
+// messageEnvelope wraps Watermill message and contains destination topic.
+type messageEnvelope struct {
+	DestinationTopic string `json:"destination_topic"`
+
+	UUID     string            `json:"uuid"`
+	Payload  []byte            `json:"payload"`
+	Metadata map[string]string `json:"metadata"`
+}
+
+func newMessageEnvelope(destTopic string, uuid string, payload []byte, metadata map[string]string) (*messageEnvelope, error) {
```
roblaszczak, line 20: maybe you can pass watermill message instead of X params? ;)

components/forwarder/envelope.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	return e, nil
+}
+
+func (e *messageEnvelope) validate() error {
+	if e.DestinationTopic == "" {
+		return errors.New("unknown destination topic")
+	}
+	if e.UUID == "" {
```
roblaszczak, line 39: well, in theory, you can have watermill message with empty UUID ;)

czeslavo: Is it correct scenario then? I don't know what would that mean.

components/forwarder/envelope.go
czeslavo: True: 40e39a7

components/forwarder/envelope.go
roblaszczak: yes, this is correct scenario

czeslavo: 00a0ad8

components/forwarder/forwarder.go
```diff
@@ -0,0 +1,127 @@
...
+func (f *Forwarder) DecoratePublisher(publisherIn message.Publisher) message.Publisher {
+	return NewPublisherDecorator(publisherIn, PublisherConfig{
+		ForwarderTopic: f.config.ForwarderTopic,
+	})
+}
+
+func (f *Forwarder) forwardMessage(msg *message.Message) error {
+	destTopic, unwrappedMsg, err := unwrapMessageFromEnvelope(msg)
+	if err != nil {
+		return errors.Wrap(err, "cannot unwrap message from an envelope")
```
roblaszczak, line 119: if you are returning error it will nack 4ever ;) I think that it may be a good idea to inject Watermill logger here and log error

roblaszczak: and log the payload! so in case if this msg should be recovered it will not be lost

roblaszczak: > Note: Keep in mind that by default the forwarder will nack all messages which weren't sent using a decorated publisher.

ah, like that, so maybe it will be a good idea to have boolean in config to encapsulate adding middleware?

components/forwarder/forwarder_test.go
```diff
@@ -0,0 +1,176 @@
...
+		require.NoError(t, publisherOut.Close())
+		require.NoError(t, subscriberOut.Close())
+	}()
+
+	// Create test context with a 5 seconds timeout so it will close any subscriptions/handlers running in the background
+	// in case of too long test execution.
+	ctx, cancelCtx := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancelCtx()
+
+	messageForwardedDetector, messageForwardedCh := setupMessageForwardedDetectorMiddleware()
```
roblaszczak, line 40: maybe using GoChannel Pub/Sub will be a bit better test? it will be more "end to end"

roblaszczak: so in other words, it may work like `gochannel Pub/Sub 1 -> forwarder -> gochannel Pub/Sub 2`

components/forwarder/forwarder_test.go
```diff
@@ -0,0 +1,176 @@
...
+	logger = watermill.NewStdLogger(true, true)
+
+	forwarderTopic = "forwarder_topic"
+	outTopic       = "out_topic"
+)
+
+// TestForwarder tests forwarding messages from PubSubIn to PubSubOut (which are GoChannel implementation underneath).
+func TestForwarder(t *testing.T) {
+	// Create a set of publisher and subscribers for both In and Out Pub/Subs.
+	publisherIn, subscriberIn := newPubSubIn(logger)
```
roblaszczak, line 26: maybe it would be better to create forwarder per every test? to be sure that we don't have any side effects (we can also run them in parallel in that case)

components/forwarder/forwarder_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+	return messagesCh
+}
+
+func requireFirstMessage(t *testing.T, expectedMessage *message.Message, ch <-chan *message.Message) {
+	select {
+	case receivedMessage := <-ch:
+		require.NotNil(t, receivedMessage)
+
+		assert.Equal(t, expectedMessage.UUID, receivedMessage.UUID)
```
roblaszczak, line 168: probably it's a better idea to use `message.Equals` here (for the forward compatibility)

components/forwarder/publisher.go
```diff
@@ -0,0 +1,64 @@
...
+	if c.ForwarderTopic == "" {
+		return errors.New("empty forwarder topic")
+	}
+
+	return nil
+}
+
+// PublisherDecorator changes `Publish` method behavior so it wraps a sent message in an envelope
+// and sends it to the forwarder topic provided in the config.
+type PublisherDecorator struct {
```
roblaszczak, line 30: I don't know why, but "Decorator" in this context is a bit misleading for me 🤔 maybe just forwarder.Publisher will be enough?

components/forwarder/forwarder.go
```diff
@@ -0,0 +1,127 @@
...
+	return f.router.Running()
+}
+
+// DecoratePublisher is a helper method which decorates a given publisher so when it publishes messages,
+// it envelopes them and publish directly to the forwarder's topic provided in the forwarder's config.
+//
+// Enveloping means that the message is put into the generic envelope containing the original message
+// and a destination topic taken from publisher `Publish` method. The destination topic is used later
+// by the forwarder to forward it to a specific topic.
+func (f *Forwarder) DecoratePublisher(publisherIn message.Publisher) message.Publisher {
```
roblaszczak, line 110: I feel something tricky here - from one point it may be nice to have some helper like that, but from another side, it's something wired in creating publisher from forwarder :P

components/forwarder/forwarder.go
czeslavo: > boolean in config to encapsulate adding middleware 
Sounds like a good idea :)

components/forwarder/forwarder_test.go
czeslavo: Actually it's already like that. `newPubSubOut/In` create GoChannel Pub/Subs. Maybe it's misleading that I need this middleware. The middleware is helpful when I want to detect if the message was nacked (to not wait some fixed duration for NOT receiving the message on out topic.

components/forwarder/forwarder_test.go
czeslavo: 7154e00

components/forwarder/forwarder_test.go
roblaszczak: ah right, hmm I still feel lack of checking what "end subscriber" receive, so let's imagine that we will use router in the wrong way and message will be not forwarded properly after middleware

components/forwarder/forwarder.go
czeslavo: 71f6a78

components/forwarder/forwarder_test.go
czeslavo: ef8223a

components/forwarder/publisher.go
czeslavo: Yeah, I agree, 328724a

components/forwarder/forwarder.go
czeslavo: I thought it may be handy to have something like this, but I'm ok with dropping it. In the end it's better to expose as narrow API as possible. And still this can be achieved with `NewPublisher` func with no troubles.

czeslavo: c71e789

roblaszczak: > narrow API as possible
yeah, this

components/forwarder/forwarder_test.go
czeslavo: Good idea: f72ba48

PR #204, opened 2020-10-27
Accept a Registerer instead of a Registry in PrometheusMetricsBuilder

sagikazarmark:
Accepting a prometheus.Registerer allows using prometheus.DefaultRegisterer.


roblaszczak:
As always, thanks! :beers:

PR #207, opened 2020-11-11
Update README.md - broken links

kochetkov-av:
Fix two broken links


m110:
Thank you @kochetkov-av!

PR #209, opened 2020-11-14
Add Server-Sent Events example

m110:


_examples/real-world-examples/server-sent-events/server/event_handlers.go
```diff
@@ -0,0 +1,167 @@
...
+) (message.Publisher, message.Subscriber, error) {
+	router, err := message.NewRouter(message.RouterConfig{}, logger)
+	if err != nil {
+		return nil, nil, err
+	}
+	router.AddMiddleware(middleware.Recoverer)
+
+	pubsub := gochannel.NewGoChannel(gochannel.Config{}, logger)
+
+	publishEvents := func(ctx context.Context, tags []string) (messages []*message.Message, err error) {
```
roblaszczak, line 32: why not create a normal function, instead of a variable?

_examples/real-world-examples/server-sent-events/server/event_handlers.go
```diff
@@ -0,0 +1,167 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			logger.Info("Adding post", watermill.LogFields{"post": event.Post})
+
+			if len(event.Post.Tags) > 0 {
+				for _, tag := range event.Post.Tags {
+					logger.Info("Adding tag", watermill.LogFields{"tag": tag})
+					err = feedsStorage.Add(msg.Context(), tag)
```
roblaszczak, line 87: by storage you mean read model? so maybe it's better to use read model name to make it more clear?

_examples/real-world-examples/server-sent-events/server/http.go
```diff
@@ -0,0 +1,318 @@
...
+	if err != nil {
+		logAndWriteError(router.Logger, w, err)
+		return
+	}
+
+	w.WriteHeader(204)
+}
+
+func (router Router) addPost(ctx context.Context, post Post) error {
+	err := router.PostsStorage.Add(ctx, post)
```
roblaszczak, line 171: maybe posts storage can be write model (BTW it may be good to explain difference between read model and write model in readme)

_examples/real-world-examples/server-sent-events/server/event_handlers.go
```diff
@@ -0,0 +1,167 @@
...
+
+				}
+
+				err = feedsStorage.AppendPost(msg.Context(), event.Post)
+				if err != nil {
+					return nil, err
+				}
+			}
+
+			return publishEvents(msg.Context(), event.Post.Tags)
```
roblaszczak, line 100: the name publishEvents is a bit missleading, it suggests that it is publishing something, and it doesn't say what events they are. maybe generate feed updated events? or better, generate read model updated events (because feed was updated earlier, read model is updated now)

_examples/real-world-examples/server-sent-events/server/event_handlers.go
```diff
@@ -0,0 +1,167 @@
...
+			newMessage := message.NewMessage(watermill.NewUUID(), payload)
+
+			messages = append(messages, newMessage)
+		}
+
+		return messages, nil
+	}
+
+	router.AddHandler(
+		"on-post-created",
```
roblaszczak, line 62: maybe something about read model update will be more concrete?

_examples/real-world-examples/server-sent-events/server/event_handlers.go
```diff
@@ -0,0 +1,167 @@
...
+		func(msg *message.Message) (messages []*message.Message, err error) {
+			defer func() {
+				if err == nil {
+					logger.Info("Event handler on-post-updated executed successfully", nil)
+				} else {
+					logger.Error("Error in on-post-updated event handler", err, nil)
+				}
+			}()
+
+			event := PostUpdated{}
```
roblaszczak, line 119: it would work nicely with CQRS...but I'm not sure if it will not shadow this example a bit? but :man_shrugging:

_examples/real-world-examples/server-sent-events/server/http.go
```diff
@@ -0,0 +1,318 @@
...
+		return nil, false
+	}
+
+	return feed, true
+}
+
+func (f feedStreamAdapter) Validate(r *http.Request, msg *message.Message) (ok bool) {
+	feedUpdated := FeedUpdated{}
+
+	err := json.Unmarshal(msg.Payload, &feedUpdated)
```
roblaszczak, line 257: it just came to my mind that metadata may be nice here as well (you don't need to unmarshal) but I think that this way is a bit more obvious

_examples/real-world-examples/server-sent-events/server/models.go
```diff
@@ -0,0 +1,70 @@
+package main
+
+import (
+	"regexp"
+	"strings"
+	"time"
+)
+
+// Note that in a real application using both "json" and "bson" tags in the same structure is strongly discouraged.
+// We use common models for database storage and HTTP API just to make this example simple and easy to grasp.
```
roblaszczak, line 10: maybe link to DRY art?

_examples/real-world-examples/server-sent-events/server/event_handlers.go
maclav3: also, this does not seem to publish any messages, just return them - so the name doesn't fit

_examples/real-world-examples/server-sent-events/server/feeds_storage.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func (s FeedsStorage) appendPostIfNotPresent(ctx context.Context, post Post) error {
+	if len(post.Tags) == 0 {
+		return nil
+	}
+
+	filter := bson.M{
+		"_id": bson.M{
+			"$in": post.Tags,
```
maclav3, line 157: are `post.Tags` a slice of mongo `_id`s? or is the `_id` field not a mongo `_id`, but any string?

m110: The feed's name is used as `_id`, and tags are feed names. It's just a string, I didn't bother with Mongo's `ObjectId` for simplicity's sake.

_examples/real-world-examples/server-sent-events/server/event_handlers.go
m110: Seems like a quick idea I did one year ago and didn't refactor properly. 😅 

c4eba92ae51e6cb3656a235ee014acfb11293212

_examples/real-world-examples/server-sent-events/server/event_handlers.go
m110: c4eba92ae51e6cb3656a235ee014acfb11293212

_examples/real-world-examples/server-sent-events/server/event_handlers.go
m110: Yeah, I'm afraid CQRS might be too much for this example, as it may get hard to follow.

_examples/real-world-examples/server-sent-events/server/event_handlers.go
m110: c4eba92ae51e6cb3656a235ee014acfb11293212

_examples/real-world-examples/server-sent-events/server/http.go
m110: It would be simpler from the code execution point of view, but I think it's easier to understand "real" events being published.

_examples/real-world-examples/server-sent-events/server/models.go
m110: c4eba92ae51e6cb3656a235ee014acfb11293212

_examples/real-world-examples/server-sent-events/server/event_handlers.go
m110: Feeds don't have a write model, they are essentially the read model for posts. I'm not sure what would be better, `postsReadModel` and `postsWriteModel` seem to not hold the information that the latter is a feed. Also, I'm worried that "model" could be confused with a struct in `models.go`. Any ideas?

_examples/real-world-examples/server-sent-events/server/http.go
m110: a378fb8e009055137f551096f86dd85b0144524a

_examples/real-world-examples/server-sent-events/server/event_handlers.go
roblaszczak: isn't mysql write model in that case?

_examples/real-world-examples/server-sent-events/server/event_handlers.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+const (
+	PostCreatedTopic = "post-created"
+	PostUpdatedTopic = "post-updated"
+	FeedUpdatedTopic = "feed-updated"
+)
+
+func SetupMessageRouter(
+	feedsStorage FeedsStorage,
```
roblaszczak, line 21: maybe it can be read model here as well?

_examples/real-world-examples/server-sent-events/server/event_handlers.go
roblaszczak: aah ok I got it now

_examples/real-world-examples/server-sent-events/server/feeds_storage.go
maclav3: fair enough

maclav3: might be confusing tho, for people expecting `_id` to be a `ObjectID`
but if it would add a lot of complication to separate those, then let's leave it

_examples/real-world-examples/server-sent-events/server/feeds_storage.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return feeds, nil
+}
+
+func (s FeedsStorage) ByName(ctx context.Context, name string) (Feed, error) {
+	filter := bson.M{
+		"_id": name,
```
maclav3, line 93: same thing here, with the `_id` an `ObjectID`.
may be confusing, but up to you

_examples/real-world-examples/server-sent-events/server/feeds_storage.go
```diff
@@ -0,0 +1,213 @@
...
+	return nil
+}
+
+func isDuplicateError(err error) bool {
+	mErr, ok := err.(mongo.WriteException)
+	if !ok {
+		return false
+	}
+
+	return mErr.WriteErrors[0].Code == 11000
```
maclav3, line 212: is there a const for that? :mage:

_examples/real-world-examples/server-sent-events/server/http.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+func (router Router) GeneratePost(w http.ResponseWriter, r *http.Request) {
+	title := gofakeit.Sentence(5)
+	content := gofakeit.Sentence(20)
+	author := gofakeit.Name()
+
+	tagsCount := rand.Intn(3) + 1
+	for i := 0; i < tagsCount; i++ {
+		content += fmt.Sprintf(" #%v", generatedTags[rand.Intn(len(generatedTags))])
```
maclav3, line 153: could use `gofakeit.RandomString` for that, since the example uses `gofakeit` already.
however, `s[rand.Intn(len(s))]` is kinda idiomatic in Golang now, so it might be even more readable :grin:

_examples/real-world-examples/server-sent-events/server/feeds_storage.go
m110: Mongo doesn't specify it has to be ObjectId, though. `_id` is just a unique primary key and indexed by default, which makes sense for the feed name.

_examples/real-world-examples/server-sent-events/server/http.go
m110: I wanted to use a static list for these tags, as `RandomString` would produce a new string every time. In the current approach, you can see the posts in feed number increase as you click the button.

_examples/real-world-examples/server-sent-events/server/feeds_storage.go
m110: There's not. 😢

roblaszczak (approved):

PR #212, opened 2020-12-07
_example/metrics: bump watermill version to latest

xyluet:
Bump watermill version to latest on _example/metrics. The old version is using RC version, that wrongly attaching promhttp handler instead of chi router handler.


m110:
Thank you @xyluet!

PR #220, opened 2021-02-26
Add `mill googlecloud subscription ls`

maclav3:
Adds a command `mill googlecloud subscription ls` that lists googlecloud subscriptions.

A topic may be given to filter the results.


tools/mill/cmd/googlecloud.go
```diff
@@ -256,6 +278,120 @@ func removeSubscription(id string) error {
...
+
+		err = printSubscriptionInfo(name, config)
+		if err != nil {
+			return errors.Wrapf(err, "error printing subscription '%s'", name)
+		}
+	}
+}
+
+func printSubscriptionInfo(name string, config pubsub.SubscriptionConfig) error {
+	b, err := yaml.Marshal(subscriptionConfig{
```
roblaszczak, line 361: why YAML actually? maybe some ASCII table would be a bit better? 🤔

maclav3: I chose YAML because it's more readable than JSON.
asciitable should be OK as well.

It will be adding a new dependency, though. We already depend on yaml for some reason.

roblaszczak:
Ok, LGTM. Thanks for the contribution 🍻

PR #221, opened 2021-02-28
Implement FanIn

0michalsokolowski0:
Create component multiplexing messages from many topics on one


pubsub/gochannel/fanin.go
```diff
@@ -0,0 +1,139 @@
...
+	return f.internalRouter.Run(ctx)
+}
+
+// Running is closed when FanIn is running.
+func (f *FanIn) Running() chan struct{} {
+	return f.internalRouter.Running()
+}
+
+// Subscribe implements the standard Subscriber interface.
+func (f *FanIn) Subscribe(ctx context.Context, topic string) (<-chan *message.Message, error) {
```
m110, line 109: Does this method still make sense for `FanIn`? It seems unused now.

pubsub/gochannel/fanin.go
```diff
@@ -0,0 +1,139 @@
+package gochannel
```
roblaszczak, line 1: As long as it's not specific for a gochannel, I think that it may be worth to move to https://github.com/ThreeDotsLabs/watermill/tree/master/components :wink:

pubsub/gochannel/fanin.go
```diff
@@ -0,0 +1,139 @@
...
+		logger:         logger,
+		subscriber:     subscriber,
+		subscriptions:  map[string]struct{}{},
+	}, nil
+}
+
+// AddSubscription adds an internal subscriptions.
+// You need to call this method with slice of `fromTopics` and `toTopic` before the FanIn is started.
+// AddSubscription is idempotent.
+func (f *FanIn) AddSubscription(fromTopics []string, toTopic string) {
```
roblaszczak, line 64: > You need to call this method with slice of `fromTopics` and `toTopic` before the FanIn is started.
so maybe it's better to add it to the constructor (and add some kind of config). In that case design would be simpler :wink:

pubsub/gochannel/fanin.go
roblaszczak: this also came to my mind, how it should be used :thinking:

roblaszczak: Ah oh, I got it.

Maybe it would be better to not expose subscriber here, but just provide higher level API that is doing that under the hood? In other words - to be able to use FanIn without creating the router. So in that case I could just do

```
fanin, err := gochannel.NewFanIn(upstreamPubSub, config, logger)
// ...
fanin.Run()
```

In that case API may be much simpler to use. But the question is if in that case it fits to your use case ;-)

roblaszczak:
As it's still in progress I converted it to draft.

roblaszczak:
Hey @0michalsokolowski0, sorry for a bit of delay with the review :sweat_smile:

All good, thanks for the contribution :beers:

Issue #126, opened 2019-09-16
PostgreSQL LISTEN/NOTIFY Subscriber


roblaszczak:
The same as #5, but for PostgreSQL :)

This implementation should be added in https://github.com/ThreeDotsLabs/watermill-sql/ repository. It **need to be compatible with already existing SQL Subscriber**. 

That means, that you should use SQL publisher for publishing messages, and binlog subscriber to read them.
We would, of course, keep already existing SQL based subscriber.

I'm not expert in PostgreSQL area, so any comments and ideas are welcome! :wink:

dalu:
note that pg has a limit of about 8MB for the payload. So if you insert or update a database larger than ~8MB it errors out. That limit is also hardcoded.

breml:
I looked into this topic myself for a similar idea (SQL backed event bus) and compared my ideas with the current implementation of watermill-sql. In my understanding, the benefits in the current implementation with LISTEN/NOTIFY wouldn't be this big, due to the guarantees currently given by the sql driver (guaranteed order of events, strictly once delivery). In my understanding the only thing that could be improved by LISTEN/NOTIFY in this case would be the polling would no longer be necessary, because the subscribers would be notified about new events.

So in the case of the benchmarks I would not expect an improvement at all, because there is no timeout/sleep/backoff as long as events are read and this is the case for the complete benchmark, because all the events are already present in the database before the read test is actually started (so the LISTEN/NOTIFY would again have no impact in this case).

In my opinion, LISTEN/NOTIFY would become interesting when some of the above mentioned guarantees would be softened, such that multiple readers in the same consumer group would actually be able to process events in parallel. For this to work, an additional table listing all message uuid per consumer group would be necessary, where the subscriber could read/update with `SKIP LOOKED` (or advisory locks).

f466162:
We're just evaluating watermill for a project based on PGSQL. We found this issue and we're curious if this topic is hard to be added to watermill or if there is no need seen within the watermill project? Using PGSQLs LISTEN/NOTIFY event mechanism is great for cases where running Kafka/NATS cluster for streaming is just too expensive.

Just let me know if you made any investigations in the past. I'd rather prefer to contribute to watermill instead of reinventing a watermill-like approach for our current project.

PR #225, opened 2021-03-10
Initialize subscribers for each consumer group

breml:
While working on an implementation for #126 (PostgreSQL driver with `SKIP LOCKED`), I found it necessary to initialize for each subscriber group. Therefore I needed to add the `SubscribeInitialize` in `generateConsumerGroup`, which is used by the test `TestConsumerGroups`.


roblaszczak:
LGTM. Thanks for your contribution, @breml!

PR #229, opened 2021-04-22
Return the handler instance created within AddNoPublisherHandler

terev:
The handler instance is currently returned from `AddHandler` but not `AddNoPublisherHandler` . This means handlers created by `AddNoPublisherHandler` cant have handler specific middlewares added. I couldn't tell if this was by design or not. This should be a non breaking change.


roblaszczak:
@terev thanks for noticing that. It should work as you did in the PR :) Thanks for the contribution!

PR #246, opened 2021-10-29
Mark consumer groups as supported

boreq:
Consumer groups are currently marked as supported here:
 https://github.com/ThreeDotsLabs/watermill-firestore/blob/master/pkg/firestore/pubsub_test.go#L53


roblaszczak:
:+1: thanks!

PR #256, opened 2021-12-03
Fix a blocking subscriber prevents other subscribers from receiving a message

### LINKED ISSUE #247: [gochannel] subscribers can block each other
Hi,

I'm using `gochannel` (watermill v1.1.0) for passing messages between publishers and subscribers. It seems like that misbehaving subscriber can block other subscribers. For example, let's take a look at the following simple test - 
```golang
func TestBadSubscriberBlocksGoodSubscriber(t *testing.T) {                                                                                                                                                         
    ctx, cancel := context.WithCancel(context.Background())                                                                                                                                                        
    defer cancel()                                                                                                                                                                                                 
                                                                                                                                                                                                                   
    topic := "my.topic"                                                                                                                                                                                            
    pubsub := gochannel.NewGoChannel(                                                                                                                                                                              
        gochannel.Config{                                                                                                                                                                                          
            BlockPublishUntilSubscriberAck: false,                                                                                                                                                                 
        },                                                                                                                                                                                                         
        watermill.NewStdLoggerWithOut(ioutil.Discard, false, false),                                                                                                                                               
    )                                                                                                                                                                                                              
                                                                                                                                                                                                                   
    // the 1st subscriber is not going to receive and ack the received message                                                                                                                                     
    pubsub.Subscribe(ctx, topic)                                                                                                                                                                                        
                                                                                                                                                                                                                   
    // the 2nd subscriber is expected to receive the message                                                                                                                                                       
    s, _ := pubsub.Subscribe(ctx, topic)                                                                                                                                                                                        
                                                                                                                                                                                                                   
    pubsub.Publish(topic, message.NewMessage(uuid.New().String(), message.Payload{}))                                                                                                                              
                                                                                                                                                                                                                   
    // inject delay to make sure that the message was published                                                                                                                                                    
    time.Sleep(1 * time.Second)                                                                                                                                                                                    
                                                                                                                                                                                                                   
    select {                                                                                                                                                                                                       
    case <-s:                                                                                                                                                                                                      
        t.Log("got message as expected")                                                                                                                                                                           
    default:                                                                                                                                                                                                       
        t.Fatal("didn't received message as expected")                                                                                                                                                             
    }                                                                                                                                                                                                              
}
```

Running the above test results in the following failure - 
```
--- FAIL: TestBadSubscriberBlocksGoodSubscriber (1.00s)
    servers_aggregate_test.go:546: didn't received message as expected
FAIL
FAIL    github.com/lightbitslabs/management/cluster-manager/service     1.009s
FAIL
```

If I remove the first call to `Subscribe()` the test passes (the subscriber receives the message).
I suspect that the root cause for the failure is in the following code (from `pubsub/gochannel/pubsub.go`) - 
```golang
    go func(subscribers []*subscriber) {                                                                                                                                                                           
        for i := range subscribers {                                                                                                                                                                               
            subscriber := subscribers[i]                                                                                                                                                                           
            subscriber.sendMessageToSubscriber(message, logFields)                                                                                                                                                 
        }                                                                                                                                                                                                          
        close(ackedBySubscribers)                                                                                                                                                                                  
    }(subscribers)
```

If the first call to `sendMessageToSubscriber()` blocks, then all remaining subscribers won't receive the message.
Is it the expected behaviour?

Thanks,
Adir

boreq:
A gochannel subscriber could prevent subscribers which were registered
afterwards from receiving a message.

Steps to reproduce:
1. Create a subscriber A subscribing to topic T.
2. Create a subscriber B subscribing to topic T.
3. Send a message on topic T.
4. Subscriber B will never receive a message as long as subscriber A
   doesn't call ack.

This pull request fixes this behaviour by distributing the messages
concurrently.

Fixes #247.


roblaszczak:
LGTM, thanks for the contribution :beers:

skast96:
Is this already in the newest version of watermill ? I am experiencing the same error, except that my subscriber A is "dead" and I need a way to unsubscribe it.

zhjdhx:
这是来自QQ邮箱的假期自动回复邮件。 您好，我现在有事，无法亲自回复您的邮件。我将在看到邮件的第一时间给您回复。

PR #266, opened 2022-01-26
Allowing to add handlers while router is already running

roblaszczak:


message/router.go
```diff
@@ -317,67 +329,108 @@ func (r *Router) Run(ctx context.Context) (err error) {
...
+
+	return nil
+}
+
+// RunHandlers runs all handlers that were added after Run().
+// RunHandlers is idempotent, so can be called multiple times safely.
+func (r *Router) RunHandlers(ctx context.Context) error {
+	r.handlersLock.Lock()
+	defer r.handlersLock.Unlock()
+
```
m110, line 359: Should we check if the router is running at this point?

roblaszczak: Good point :+1:. Fixed in a756811

PR #269, opened 2022-02-19
Changes to Pubsub Test for ExactlyOnce

AlexCuse:
Noticed one bad feature check (skipping if Persistent enabled, though message indicates ExactlyOnce intended).

Also added another to skip TestMessageCtx - this can also be addressed by calling publish the second time with a new message ID but I am not sure if that would invalidate the test in other situation?

### DIFF
--- a/pubsub/tests/test_pubsub.go
+++ b/pubsub/tests/test_pubsub.go
@@ -886,6 +886,13 @@ func TestMessageCtx(
 	tCtx TestContext,
 	pubSubConstructor PubSubConstructor,
 ) {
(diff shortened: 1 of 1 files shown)


AlexCuse:
With these couple changes I am able to get a basic exactly once test running with a WIP NATS jetstream client - it makes sense to me why the changes would be needed for the NATs implementation but I am not sure if this coverage would be needed for other pubsubs that support exactly once, may need to find a different way to test.

roblaszczak:
LGTM, thanks for the contribution! @AlexCuse

PR #276, opened 2022-03-23
Fix #275

### LINKED ISSUE #275: GoChannel subsciber blocked without Acking messages
In the current implementation, the goroutine executing `sendMessageToSubscriber()` is blocked until either the message is acked/nacked or the subsciber is closing.

I will submit a PR soon.

upils:
This PR contains a fix for the issue but for now other PubSub tests are broken (eg. TestResendOnError).

I may have misunderstood how the Ack/Nack mechanism is supposed to work. 

@roblaszczak @m110 I would love to discuss it further if you have some time.

### DIFF
--- a/pubsub/gochannel/pubsub.go
+++ b/pubsub/gochannel/pubsub.go
@@ -154,7 +154,7 @@ func (g *GoChannel) sendMessage(topic string, message *message.Message) (<-chan 
 
 			wg.Add(1)
 			go func() {
-				subscriber.sendMessageToSubscriber(message, logFields)
+				subscriber.sendMessageToSubscriber(message, logFields, g.config.BlockPublishUntilSubscriberAck)
 				wg.Done()
 			}()
 		}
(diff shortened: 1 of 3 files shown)


pubsub/gochannel/pubsub_test.go
```diff
@@ -108,6 +108,72 @@ func TestPublishSubscribe_block_until_ack(t *testing.T) {
 	}
 }
 
+func TestPublishSubscribe_do_not_block_without_ack_required(t *testing.T) {
```
prashant3286, line 111: Can we use a small function name to make it more concise?

prashant3286:
Please resolve the test :)

pubsub/gochannel/pubsub.go
```diff
@@ -370,6 +370,11 @@ SendToSubscriber:
 			return
 		}
 
+		if !blockPublishUntilSubscriberAck {
+			s.logger.Trace("Sent message to subscriber without ack", logFields)
+			return
+		}
+
```
m110, line 377: The select below isn't really blocking execution, as it's running in another goroutine. `BlockPublishUntilSubscriberAck` is checked in `Publish` and that's where it blocks.

Can you share more about your use case?

Tochemey: @m110 the current implementation will cause deadlock on all subscribers even if one of them did Ack/Nack a message. However I believe only the subscriber that does not Ack/Nack should lose the message leaving the rest processing goes through. So the setting to block publishing will deadlock if one subscriber does not Ack and I don't think that the intention.

Tochemey: My suggestion will be to timeout when the `BlockPublishUntilSubscriberAck` is set to avoid the deadlock

m110:
This has been addressed in #589

Issue #253, opened 2021-11-22
A comment for the `Message.UUID` lies


xorcare:
This comment says that the message ID is optional, 
but for SQL transport it is a mandatory attribute,
in turn it causes misunderstanding? 

Is it possible to fix it or did I get something wrong?

https://github.com/ThreeDotsLabs/watermill/blob/b9928e750ba673cf93d442db88efc04706f67388/message/message.go#L20
https://github.com/ThreeDotsLabs/watermill/blob/b9928e750ba673cf93d442db88efc04706f67388/message/message.go#L21

Place of use
https://github.com/ThreeDotsLabs/watermill-sql/blob/4f39bf82b6180ca2191c791e7cb220fff22b9255/pkg/sql/schema_adapter.go#L48
https://github.com/ThreeDotsLabs/watermill-sql/blob/4f39bf82b6180ca2191c791e7cb220fff22b9255/pkg/sql/schema_adapter_postgresql.go#L33

roblaszczak:
Hey @xorcare, good point! When this comment was written SQL Pub/Sub was not implemented yet. It would be more accurate to put info that usually it's not required, but some Pub/Sub implementation may require this field.

If you have some time feel free to post PR and update doc :)

PR #287, opened 2022-06-14
Update comment for field Message.UUID #253

xorcare:
This is necessary to inform the developer that in some use cases this
field may be required.


m110:
Hey @xorcare, can you share what use cases you have in mind?

Issue #288, opened 2022-06-20
Allow router plugins access for router handlers


dangerousplay:
Hi, we are creating a middleware and also a router plugin that allows reprocessing of messages that failed and went through a poison/dead letter topic.

In order to be able to reprocess the messages without needing to republish them in the same topic, 
that other applications can be listening, the router plugin should be able to access the handlers.

We can create a router plugin that could expose the reprocess functionality via an HTTP endpoint, for example.

PR #289, opened 2022-06-20
[Added] Router handlers method for router plugins

dangerousplay:
Exposed the registered router handlers to be used by a router plugin in order to allow use cases such as message reprocessing described in issue #288


m110:
Thank you @dangerousplay! 🙌

PR #295, opened 2022-08-24
Fix #275 and add gochannel concurreny control

### LINKED ISSUE #275: GoChannel subsciber blocked without Acking messages

woostundy:
Removed the sending lock, and replaced it with a channel to control concurrency.

### DIFF
--- a/pubsub/gochannel/pubsub.go
+++ b/pubsub/gochannel/pubsub.go
@@ -380,7 +380,7 @@ SendToSubscriber:
 			s.logger.Trace("Message acked", logFields)
 			return
 		case <-msgToSend.Nacked():
-			s.logger.Trace("Nack received", logFields)
+			s.logger.Trace("Nack received, resending message", logFields)
 			continue SendToSubscriber
 		case <-s.closing:
 			s.logger.Trace("Closing, message discarded", logFields)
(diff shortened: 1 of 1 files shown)


woostundy:
#275

pubsub/gochannel/pubsub.go
```diff
@@ -342,12 +346,13 @@ func (s *subscriber) Close() {
 }
 
 func (s *subscriber) sendMessageToSubscriber(msg *message.Message, logFields watermill.LogFields) {
-	s.sending.Lock()
```
m110, line 345: We might still need the lock to avoid race conditions with `Close`.

pubsub/gochannel/pubsub.go
```diff
@@ -342,12 +346,13 @@ func (s *subscriber) Close() {
...
-	defer s.sending.Unlock()
+	s.sendingChannel <- struct{}{}
+	defer func() {
+		<-s.sendingChannel
```
m110, line 351: `sendMessageToSubscriber` is also called to deliver persisted messages to newly joining subscribers. I'm not sure if this will behave as expected in this case. Can we have some tests to cover it?

woostundy: Sure, I'll check again and add some tests for it.

Issue #235, opened 2021-07-31
Router does not wait for running handlers to complete.


checkmunza:
I can't find anywhere that the `Router` will call `Wait` method on `runningHandlersWg`.
```go
type Router struct {
        ...
        
	handlersWg        *sync.WaitGroup
	runningHandlersWg *sync.WaitGroup
        
        ...
}
```

Should it need to do that so we can close the router gracefully?

```go
// Close gracefully closes the router with a timeout provided in the configuration.
func (r *Router) Close() error {
	r.closedLock.Lock()
	defer r.closedLock.Unlock()

	if r.closed {
		return nil
	}
	r.closed = true

	r.logger.Info("Closing router", nil)
	defer r.logger.Info("Router closed", nil)

	close(r.closeCh)
	defer close(r.closedCh)

	timeouted := sync_internal.WaitGroupTimeout(r.handlersWg, r.config.CloseTimeout)
	if timeouted {
		return errors.New("router close timeouted")
	}

	return nil
}
```

anandsunderraman:
Yeah I guess I am also facing a similar issue.
When I shutdown my program I get this error message that says
```
level=ERROR msg="All handlers stopped, closing router" err="all router handlers stopped" 
```

roblaszczak:
Fixed here: https://github.com/ThreeDotsLabs/watermill/pull/304

PR #304, opened 2022-10-05
Make router wait for running handlers before shutdown

Eittipat-KKP:
Fix for issue https://github.com/ThreeDotsLabs/watermill/issues/235


roblaszczak:
@Eittipat-KKP, thanks for the contribution!

PR #306, opened 2022-11-01
Update pubsub.go

golang-boy:
no subscriber should return an error when publishing


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/ThreeDotsLabs/watermill
+module github.com/golang-boy/watermill
```
m110, line 1:
```suggestion
module github.com/ThreeDotsLabs/watermill
```

pubsub/gochannel/pubsub.go
```diff
@@ -142,8 +142,9 @@ func (g *GoChannel) sendMessage(topic string, message *message.Message) (<-chan
 
 	if len(subscribers) == 0 {
 		close(ackedBySubscribers)
-		g.logger.Info("No subscribers to send message", logFields)
-		return ackedBySubscribers, nil
+		err := errors.New("No subscribers")
+		g.logger.Error("No subscribers to send message", err, logFields)
+		return ackedBySubscribers, err
```
m110, line 147: There's no error by design, as `Publish` drops the message if no subscribers are waiting for it. If you'd like to see an error in this case, we would need a configuration option not to change the current behavior.

PR #312, opened 2022-11-17
example > basic > 3-router: reorder middlewares

benshields:
Looking at the source code & examples I see that middlewares are executed top to bottom. In the comments here we see that Recoverer should happen before Retry. This should help folks that look at these examples.

### DIFF
--- a/_examples/basic/3-router/main.go
+++ b/_examples/basic/3-router/main.go
@@ -32,8 +32,9 @@ func main() {
 
 	// Router level middleware are executed for every message sent to the router
 	router.AddMiddleware(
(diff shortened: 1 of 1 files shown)


m110:
Hey @benshields. The current code is correct. The retry middleware "wraps" the handler, so when it is executed first, it goes straight to the next middleware and then checks the error.

To see this in action, add a `panic` in the example's handler. To better see the effect, increase the timeout interval to a few seconds. You will see the panics turned into errors, then passed to the retry logic. When you change the middleware order like in the PR, the retry middleware won't kick in (since a panic is raised).

PR #327, opened 2023-01-24
Fixing a few small typos

gnagel:
Nothing fancy, just a few typos that I noticed while reading the code


m110:
Thank you @gnagel :)

PR #333, opened 2023-02-03
Add generic jsonendpoint adapter

dkotik:
Package jsonendpoint provides a generic adapter that converts a validated JSON request into a [message.Message] published to a certain topic. Use together with an HTTP router to build gateways to any [message.Publisher].

Had to bump Golang version to `1.18` in the `go.mod` file to support generics.

This is my first contribution.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/ThreeDotsLabs/watermill
 
-go 1.17
+go 1.18
 
 require (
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ThreeDotsLabs/watermill
 
-go 1.17
+go 1.18
```
dkotik, line 3: Required for generics support

PR #356, opened 2023-03-14
added support for Event Handler groups

roblaszczak:


components/cqrs/command_bus.go
```diff
@@ -3,18 +3,21 @@ package cqrs
...
 )
 
 // CommandBus transports commands to command handlers.
 type CommandBus struct {
-	publisher     message.Publisher
-	generateTopic func(commandName string) string
-	marshaler     CommandEventMarshaler
+	publisher message.Publisher
+
+	config CommandProcessorConfig
```
maclav3, line 16: the struct is named `CommandBus`, but the config is named `CommandProcessorConfig`
is taht ok?

roblaszczak:
The implementation was ported in a fresh PR: https://github.com/ThreeDotsLabs/watermill/pull/367

components/cqrs/command_bus.go
roblaszczak: yeah, it was a leftover 😅 but it was already re-written ten times here: https://github.com/ThreeDotsLabs/watermill/pull/367 (and should be fine now)

PR #366, opened 2023-06-05
Added request-reply support for commands

roblaszczak:


AlexCuse:
This is awesome I've stitched together an ad-hoc equivalent to this at least a few times be nice to have something available in package.

AlexCuse:
Been thinking about this a lot you may want to drop the CQRS namespace since request / reply is generally useful in a lot of situations.  I don't have a great name in mind though.  We built it into the message bus client for edgexfoundry mainly for the performance benefits that come with sending command/control messages to local device services without using HTTP.  

The messagebus implementation is not really too much different than a composite of a WM publisher and subscriber.  Its super easy to adapt between the two.  In case you want to look could start here https://github.com/edgexfoundry/go-mod-messaging/blob/main/internal/pkg/request.go

qiulin:
Is there any process?

m110:
Implemented in #397.

PR #372, opened 2023-07-15
fix: router new logger panic

vladtenlive:
Initialization with `nil logger` will cause **panic** on Router run 
https://github.com/vladtenlive/watermill/blob/3867ad50830e3b50d13685243c68c331cd42ff32/message/router.go#L355

image


m110 (approved):

m110:
Thank you @vladtenlive!

PR #387, opened 2023-09-05
Forwarder: propagate the enveloped message's Context

m110:
This may be useful if the Forwarder's router middleware changes the Context in some way.


components/forwarder/envelope.go
```diff
@@ -69,6 +69,7 @@ func unwrapMessageFromEnvelope(msg *message.Message) (destinationTopic string, u
 
 	watermillMessage := message.NewMessage(envelopedMsg.UUID, envelopedMsg.Payload)
 	watermillMessage.Metadata = envelopedMsg.Metadata
+	watermillMessage.SetContext(envelopedMsg.Context())
```
m110, line 72:
```suggestion
	watermillMessage.SetContext(msg.Context())
```

Issue #392, opened 2023-09-14
Allow for custom labels on prometheus metrics


matdurand:
As discussed on Discord, it would be nice to be able to configure custom labels on prometheus metrics.

PR #393, opened 2023-09-14
Allow for custom labels on prometheus metrics

matdurand:
Solves https://github.com/ThreeDotsLabs/watermill/issues/392

### DIFF
--- a/components/metrics/builder.go
+++ b/components/metrics/builder.go
@@ -7,12 +7,28 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-func NewPrometheusMetricsBuilder(prometheusRegistry prometheus.Registerer, namespace string, subsystem string) PrometheusMetricsBuilder {
-	return PrometheusMetricsBuilder{
+type Option func(b *PrometheusMetricsBuilder)
+
+// CustomLabel allow to provide a custom metric label and a function to compute the value
+func CustomLabel(label string, fn LabelComputeFn) Option {
+	return func(b *PrometheusMetricsBuilder) {
+		b.customLabels = append(b.customLabels, metricLabel{
+			label:     label,
+			computeFn: fn,
+		})
+	}
+}
+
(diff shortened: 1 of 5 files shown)


components/metrics/builder.go
```diff
@@ -7,12 +7,28 @@ import (
...
+func CustomLabel(label string, fn LabelComputeFn) Option {
+	return func(b *PrometheusMetricsBuilder) {
+		b.customLabels = append(b.customLabels, metricLabel{
+			label:     label,
+			computeFn: fn,
+		})
+	}
+}
+
+func NewPrometheusMetricsBuilder(prometheusRegistry prometheus.Registerer, namespace string, subsystem string, opts ...Option) PrometheusMetricsBuilder {
```
m110, line 22: Hey @matdurand, sorry about the delay. I wanted to comment last week and just realized I didn't do it. 🤦‍♂️ 

I have just one comment regarding the API — most Watermill components use a "Config" struct instead of the Option pattern, and I feel it would be a good idea to be consistent, just for the improved developer experience. This means we would need another constructor, something like `NewPrometheusMetricsBuilderWithConfig`. What do you think?

matdurand: Hey @m110, yeah sure, consistency is paramount. I replaced Option with config. I also renamed "custom" for "additional", just because it sounded better imo. Let me know if these changes are enough. Thanks

components/metrics/labels.go
```diff
@@ -46,3 +46,30 @@ func labelsFromCtx(ctx context.Context, labels ...string) prometheus.Labels {
 
 	return ctxLabels
 }
+
+type LabelComputeFn func(msgCtx context.Context) string
```
matdurand, line 50: We might want to reflect on that signature. Is it enough to use only the context to get a label's value? Should we also pass the message pointer? For my use case, the additional labels are static, so I basically do this
```
        metrics.MetricLabel{
		Label: "service",
		ComputeFn: func(ctx context.Context) string {
			return "my_service"
		},
	},
```
but there might be a use case for getting something from the message itself as a label value.

roblaszczak (approved):

PR #397, opened 2023-09-20
Request/reply support

roblaszczak:


components/requestreply/requestreply.go
```diff
@@ -0,0 +1,91 @@
...
+	// If listening for reply timed out or context was canceled, Error is ReplyTimeoutError.
+	//
+	// If error from handler is returned, CommandHandlerError is returned.
+	// If processing was successful, Error is nil.
+	Error error
+
+	// NotificationMessage contains the notification message send after command is handled.
+	// It's only present if the request/reply backend is using Pub/Sub for notifications (for example: PubSubBackend).
+	//
+	// Warning: NotificationMessage is nil if timeout occurred.
```
m110, line 32:
```suggestion
	// HandlerResult contains the handler result.
	// It's preset only when NewCommandHandlerWithResult is used. If NewCommandHandler is used, HandlerResult is empty.
	//
	// Result is sent even if the handler returns an error.
	HandlerResult Result

	// Error contains the error returned by the command handler or the Backend when handling notification fails.
	// Handling the notification can fail, for example, when unmarshaling the message or if there's a timeout.
	// If listening for a reply times out or the context is canceled, the Error is ReplyTimeoutError.
	//
	// If an error from the handler is returned, CommandHandlerError is returned.
	// If processing was successful, Error is nil.
	Error error

	// NotificationMessage contains the notification message sent after the command is handled.
	// It's present only if the request/reply backend uses a Pub/Sub for notifications (for example, PubSubBackend).
	//
	// Warning: NotificationMessage is nil if a timeout occurs.
```

components/requestreply/requestreply.go
```diff
@@ -0,0 +1,91 @@
...
+type BackendOnCommandProcessedParams[Result any] struct {
+	Command        any
+	CommandMessage *message.Message
+
+	HandlerResult Result
+	HandleErr     error
+}
+
+// OperationID is a unique identifier of a command.
+// It's used to correlate command with reply between bus and handler.
```
m110, line 55:
```suggestion
// It's used to correlate commands with replies between the bus and the handler.
```

components/requestreply/handler.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// NewCommandHandler creates a new CommandHandler which supports request-reply pattern.
+// The result handler, is handler compatible with cqrs.CommandHandler.
+//
+// The logic if command should be acked or not is based on logic of the Backend.
+// For example, for the PubSubBackend, it depends on `PubSubBackendConfig.AckCommandErrors` option.
+//
+// The logic if command should be acked or not is based on logic of the Backend.
+// For example, for the PubSubBackend, it depends on `PubSubBackendConfig.AckCommandErrors` option.
```
m110, line 18:
```suggestion
// NewCommandHandler creates a new CommandHandler which supports the request-reply pattern.
// The result handler is handler compatible with cqrs.CommandHandler.
//
// The logic if a command should be acked or not is based on the logic of the Backend.
// For example, for the PubSubBackend, it depends on the `PubSubBackendConfig.AckCommandErrors` option.
```

components/requestreply/handler.go
```diff
@@ -0,0 +1,81 @@
...
+// NewCommandHandlerWithResult creates a new CommandHandler which supports request-reply pattern with result.
+// The result handler, is handler compatible with cqrs.CommandHandler.
+//
+// In addition to cqrs.CommandHandler, it also allows to return result from handler.
+// The result is passed to the Backend implementation and sent to the caller.
+//
+// The logic if command should be acked or not is based on logic of the Backend.
+// For example, for the PubSubBackend, it depends on `PubSubBackendConfig.AckCommandErrors` option.
+//
+// The reply is sent to the caller, even if handler returns an error.
```
m110, line 46:
```suggestion
// NewCommandHandlerWithResult creates a new CommandHandler which supports the request-reply pattern with a result.
// The result handler is handler compatible with cqrs.CommandHandler.
//
// In addition to cqrs.CommandHandler, it also allows returninga result from the handler.
// The result is passed to the Backend implementation and sent to the caller.
//
// The logic if a command should be acked or not is based on the logic of the Backend.
// For example, for the PubSubBackend, it depends on the `PubSubBackendConfig.AckCommandErrors` option.
//
// The reply is sent to the caller, even if the handler returns an error.
```

m110, line 46:
```suggestion
// NewCommandHandlerWithResult creates a new CommandHandler which supports the request-reply pattern with a result.
// The result handler is handler compatible with cqrs.CommandHandler.
//
// In addition to cqrs.CommandHandler, it also allows returninga result from the handler.
// The result is passed to the Backend implementation and sent to the caller.
//
// The logic if a command should be acked or not is based on the logic of the Backend.
// For example, for the PubSubBackend, it depends on the `PubSubBackendConfig.AckCommandErrors` option.
//
// The reply is sent to the caller, even if the handler returns an error.
```

components/requestreply/handler.go
```diff
@@ -0,0 +1,81 @@
...
+	originalMessage := cqrs.OriginalMessageFromCtx(ctx)
+	if originalMessage == nil {
+		// this should not happen, as long as cqrs.CommandProcessor is used - but it's not mandatory
+		// in this case it's enough to use cqrs.CtxWithOriginalMessage
+		return nil, errors.New(
+			"original message not found in context, did you passed context correctly everywhere? " +
+				"did you used cqrs.CommandProcessor? " +
+				"if you are using custom implementation, please call cqrs.CtxWithOriginalMessage on context passed to handler",
+		)
+	}
```
m110, line 79:
```suggestion
		// This should not happen, as long as cqrs.CommandProcessor is used - but it's not mandatory.
		// In this case, it's enough to use cqrs.CtxWithOriginalMessage
		return nil, errors.New(
			"original message not found in context, did you pass context correctly everywhere? " +
				"did you use cqrs.CommandProcessor? " +
				"if you are using custom implementation, please call cqrs.CtxWithOriginalMessage on the context passed to the handler",
		)
	}
```

components/requestreply/command_bus.go
```diff
@@ -0,0 +1,80 @@
...
+//			ts.CommandBus,
+//			ts.RequestReplyBackend,
+//			&TestCommand{ID: "1"},
+//		)
+//
+// SendWithReply will send the replies to the channel until the context is cancelled or the timeout is exceeded.
+// They are multiple cases when more than one reply can be sent:
+//   - when the handler returns an error, and backend is configured to nack the message on error
+//     (for the PubSubBackend, it depends on `PubSubBackendConfig.AckCommandErrors` option.),
+//   - when you are using fan-out mechanism and commands are handled multiple times,
```
m110, line 47:
```suggestion
// SendWithReply sends a command to the command bus and receives replies from the command handler(s).
// It returns a channel with replies, cancel function, and error.
//
// SendWithReply can be canceled by calling the cancel function, canceling the context, or exceeding the timeout set in the backend (if set).
// When SendWithReply is canceled, the returned channel is closed as well.
// 
// Warning: It's important to cancel the function because it listens for the replies in the background.
// Not canceling it can lead to subscriber leaks.
//
// SendWithReply can listen for handlers with results (NewCommandHandlerWithResult) and without results (NewCommandHandler).
// When listening for handlers without results, you should pass `NoResult` or `struct{}` as the `Result` generic type:
//
//	 replyCh, cancel, err := requestreply.SendWithReply[requestreply.NoResult](
//			context.Background(),
//			ts.CommandBus,
//			ts.RequestReplyBackend,
//			&TestCommand{ID: "1"},
//		)
//
// If the `NewCommandHandlerWithResult` handler returns a specific type, you should pass it as the `Result` generic type:
//
//	 replyCh, cancel, err := requestreply.SendWithReply[SomeTypeReturnedByHandler](
//			context.Background(),
//			ts.CommandBus,
//			ts.RequestReplyBackend,
//			&TestCommand{ID: "1"},
//		)
//
// SendWithReply will send the replies to the channel until the context is canceled or the timeout is exceeded.
//There are multiple cases when more than one reply can be sent:
//   - when the handler returns an error, and the backend is configured to nack the message on error
//     (for the PubSubBackend, it depends on the `PubSubBackendConfig.AckCommandErrors` option),
//   - when you use a fan-out mechanism and commands are handled multiple times by more than one handler).
```

components/requestreply/backend_pubsub_marshaler.go
```diff
@@ -0,0 +1,60 @@
...
+	} else {
+		msg.Metadata.Set(HasErrorMetadataKey, "0")
+	}
+
+	b, err := json.Marshal(params.HandlerResult)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot marshal reply")
+	}
+
+	msg.Metadata.Set(ResultMetadataKey, string(b))
```
m110, line 41: Consider using the payload for the results instead of the metadata.

components/requestreply/command_bus.go
```diff
@@ -0,0 +1,80 @@
...
+//			ts.RequestReplyBackend,
+//			&TestCommand{ID: "1"},
+//		)
+//
+// SendWithReply will send the replies to the channel until the context is canceled or the timeout is exceeded.
+//There are multiple cases when more than one reply can be sent:
+//   - when the handler returns an error, and the backend is configured to nack the message on error
+//     (for the PubSubBackend, it depends on the `PubSubBackendConfig.AckCommandErrors` option),
+//   - when you use a fan-out mechanism and commands are handled multiple times by more than one handler).
+func SendWithReply[Result any](
```
m110, line 48: Consider adding another func, like:

```go
SendWithReply[Result any](...) (Reply[Result, error)
```

components/requestreply/backend_pubsub.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot create request/reply notifications subscriber")
+	}
+
+	replyNotificationTopic, err := p.config.GenerateSubscribeTopic(replyContext)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot generate request/reply notifications topic")
+	}
+
+	var cancel context.CancelFunc
```
m110, line 131: Seems to be unused

m110: Nevermind 🤦‍♂️

components/cqrs/command_processor.go
```diff
@@ -45,7 +45,9 @@ type CommandProcessorConfig struct {
 	Logger watermill.LoggerAdapter
 
 	// If true, CommandProcessor will ack messages even if CommandHandler returns an error.
-	// If RequestReplyEnabled is enabled and sending reply fails, the message will be nack-ed anyway.
+	// If RequestReplyBackend is not null and sending reply fails, the message will be nack-ed anyway. todo: verify
+	// todo: test if it works properly with RequestReply (it should nack by default as well?)
+	// todo: describe that it doesn't affect RequestReplyBackend filures - sending reply failure will always nack the message
 	AckCommandHandlingErrors bool
```
m110, line 51: Let's address the TODOs here

components/requestreply/backend_pubsub.go
```diff
@@ -0,0 +1,278 @@
...
+	config    PubSubBackendConfig
+	marshaler BackendPubsubMarshaler[Result]
+}
+
+// NewPubSubBackend creates a new PubSubBackend.
+//
+// If you want to use backend together with `NewCommandHandler` (without result), you should pass `NoResult` or `struct{}` as Result type.
+func NewPubSubBackend[Result any](
+	config PubSubBackendConfig,
+	marshaler BackendPubsubMarshaler[Result],
```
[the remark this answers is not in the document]
roblaszczak: it's generic and it would force config to depend on generic type - in other words you  couldn't have a single confg instance that could be used for multiple Result types

components/cqrs/command_processor.go
```diff
@@ -314,12 +316,15 @@ func (p CommandProcessor) routerHandlerFunc(handler CommandHandler, logger water
...
+		ctx := CtxWithOriginalMessage(msg.Context(), msg)
+		msg.SetContext(ctx)
+
 		if err := p.config.Marshaler.Unmarshal(msg, cmd); err != nil {
 			return err
 		}
 
 		handle := func(params CommandProcessorOnHandleParams) (err error) {
-			return params.Handler.Handle(params.Message.Context(), params.Command)
+			return params.Handler.Handle(ctx, params.Command)
```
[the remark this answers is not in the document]
roblaszczak: as long as I understand your question: it won't be processed then - the similar thing will happen if you would not use request in HTTP middleware

components/requestreply/backend_pubsub_marshaler.go
roblaszczak: done

components/requestreply/command_bus.go
roblaszczak: done

components/cqrs/command_processor.go
roblaszczak: done

m110 (approved):

mamyn0va:
Hi @roblaszczak, do you known when the request/reply feature will be released ? Is there any related doc ?

PR #403, opened 2023-10-17
move problematic return statement out of for loop

Darkskald:
The closure produced by `addEventHandlersToRoute` has a return statement inside the for loop which leads to a termination of the function in case of a successful event handling which is not the desired behavior if multiple handlers for the same event type are registered.

### DIFF
--- a/components/cqrs/event_processor_group.go
+++ b/components/cqrs/event_processor_group.go
@@ -248,9 +248,9 @@ func (p EventGroupProcessor) routerHandlerGroupFunc(handlers []GroupEventHandler
 				logger.Debug("Error when handling event", watermill.LogFields{"err": err})
 				return err
 			}
-
-			return nil
 		}
+
+		return nil
 
 		if !p.config.AckOnUnknownEvent {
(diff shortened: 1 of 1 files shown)


components/cqrs/event_processor_group.go
```diff
@@ -248,10 +248,10 @@ func (p EventGroupProcessor) routerHandlerGroupFunc(handlers []GroupEventHandler
 				logger.Debug("Error when handling event", watermill.LogFields{"err": err})
 				return err
 			}
-
-			return nil
 		}
 
+		return nil
```
m110, line 253: The `return` statement here skips the `if` below entirely, right? Is that on purpose?

ilusharulkov:
merge plz

components/cqrs/event_processor_group.go
roblaszczak: indeed, this would be a problem, it's fixed here: https://github.com/ThreeDotsLabs/watermill/commit/9d5e2da133397fcfb6e8f00eb424df91005e9c7a

PR #406, opened 2023-10-29
ci: update setup-go and checkout to latest action versions

AlexCuse:
This should resolve node deprecation warnings when viewing action results in the github UI.


roblaszczak:
Thanks @AlexCuse!

PR #418, opened 2024-01-07
Add no subscribers fallback option to GoChannel Pub/Sub

annismckenzie:
I needed a way to listen to any message sent via the GoChannel Pub/Sub that didn't have a subscriber yet. This was mainly used for debugging but it could be useful for other purposes as well.

I made sure to add a test and left the rest of the tests untouched – the test suite is passing.

Would you be open to accepting this contribution? If so, I'd take another pass over the current implementation's comments, some would need updating (for example, `That means if you send a message to a topic to which no subscriber is subscribed, that message will be discarded.` – update: this is done ☑️).

### DIFF
--- a/pubsub/gochannel/pubsub.go
+++ b/pubsub/gochannel/pubsub.go
@@ -11,9 +11,10 @@ import (
 	"github.com/ThreeDotsLabs/watermill/message"
 )
 
-// NoSubscribersFallbackTopic is the fallback topic messages without any subscribers will be sent to.
-// This is used if the `EnableFallback` configuration option is enabled.
-const NoSubscribersFallbackTopic = "*"
+// NoSubscribersFallbackDefaultTopic is the default fallback topic messages without any subscribers
+// will be sent to – it is used if the `EnableNoSubscribersFallback` option is enabled and no
+// fallback topic is configured via the `NoSubscribersFallbackTopic` option.
+const NoSubscribersFallbackDefaultTopic = "*"
 
 // Config holds the GoChannel Pub/Sub's configuration options.
 type Config struct {
@@ -32,8 +33,13 @@ type Config struct {
 	BlockPublishUntilSubscriberAck bool
 
 	// When true, messages sent to a topic without any subscribers will be sent to the
-	// subscribers of the `*` topic.
-	EnableFallback bool
+	// subscribers of the fallback topic (configured via `NoSubscribersFallbackTopic` option).
+	EnableNoSubscribersFallback bool
+
+	// NoSubscribersFallbackTopic is the fallback topic messages without any subscribers will be sent to.
+	// This is used if the `EnableNoSubscribersFallback` configuration option is enabled.
+	// If it's not set then `*` is used by default.
+	NoSubscribersFallbackTopic string
 }
 
 // GoChannel is the simplest Pub/Sub implementation.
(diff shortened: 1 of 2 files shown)


pubsub/gochannel/pubsub.go
```diff
@@ -11,6 +11,10 @@ import (
 	"github.com/ThreeDotsLabs/watermill/message"
 )
 
+// NoSubscribersFallbackTopic is the fallback topic messages without any subscribers will be sent to.
+// This is used if the `EnableFallback` configuration option is enabled.
+const NoSubscribersFallbackTopic = "*"
```
roblaszczak, line 16: what do you think about adding configuration for the topic name instead? so we could have two options in `Config`

`EnableFallback`
`NoSubscribersFallbackTopic`

what do you think?

pubsub/gochannel/pubsub.go
```diff
@@ -26,6 +30,10 @@ type Config struct {
 	// When true, Publish will block until subscriber Ack's the message.
 	// If there are no subscribers, Publish will not block (also when Persistent is true).
 	BlockPublishUntilSubscriberAck bool
+
+	// When true, messages sent to a topic without any subscribers will be sent to the
+	// subscribers of the `*` topic.
+	EnableFallback bool
```
roblaszczak, line 36: maybe it would be a bit longer, IMO it could be a bit more explicit, like `EnableNoSubscribersFallback`

roblaszczak, line 36: also, food for thought, but I wonder if this fallback couldn't be a closure 🤔

like `NoSubscribersFallback func(msg message.Message, pubsub *GoChannel) error `

It could be much more flexible 🤔 WDYT?

annismckenzie: That sounds worthwhile to investigate.

annismckenzie:
Just saw your review – thank you! I’ll take a closer look next week.

pubsub/gochannel/pubsub.go
annismckenzie: Yup, I agree that the longer version is more expressive and changed it.

pubsub/gochannel/pubsub.go
annismckenzie: I added both `EnableNoSubscribersFallback` (as per your review comment below) and `NoSubscribersFallbackTopic` to the configuration. I did want to leave the default topic (`*`) as a const in case the user doesn't set a custom fallback topic. I also added another test case for this.

pubsub/gochannel/pubsub.go
annismckenzie: I investigated this with a private project of mine where I'm using the fork and the closure is awkward to use because the function you want to call has to be known and available when instantiating the `GoChannel`. However, I'm passing this `GoChannel` object around for subscribing and publishing and then just use

```go
noSubscribersMessages, err := pubSub.Subscribe(ctx, gochannel.NoSubscribersFallbackDefaultTopic)
if err != nil {
	return fmt.Errorf("subscribing to fallback topic: %w", err)
}
go func(ctx context.Context, messages <-chan *message.Message) {
	for {
		select {
		case msg := <-messages:
			// handle message

			// lastly:
			msg.Ack()
		case <-ctx.Done():
			return
		}
	}
}(ctx, noSubscribersMessages)
```

which works well and is easy to use. That is to say, I'd leave this as is.

annismckenzie:
Short update: I rebased and addressed your comments, see above. I also updated a couple comments as promised in the pull request description in 8808fd72d6f154d234c5af4ac281a946c0b7234e and this would be ready for another look.

m110:
Hey @annismckenzie. After discussing this, we decided it's probably too specific to be included in the generic GoChannel implementation. If you still need this behavior, I think the best way would be to create your own GoChannel implementation based on the existing one. Sorry it took a while!

Thank you for contributing!

PR #419, opened 2024-01-22
Examples: update watermill-sql to v3

thpk:
While trying to figure out how to use watermill-sql, it occurred that the newest version is v3, and the examples use v2.

This updates all occurrences of watermill-sql/v2 in the examples with v3.


roblaszczak:
Good job @thpk, thanks!

PR #423, opened 2024-02-26
Add deduplication middleware and publisher decorator

dkotik:
Useful for many projects. If similar messages appear, discards all but the first one in a given time window.

### DIFF
--- a/message/router/middleware/deduplicator.go
+++ b/message/router/middleware/deduplicator.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
-	"encoding/hex"
 	"fmt"
 	"hash/adler32"
 	"io"
@@ -45,7 +44,7 @@ func NewMessageHasherAdler32(readLimit int64) MessageHasher {
 		if err != nil && err != io.EOF {
 			return "", err
 		}
-		return hex.EncodeToString(h.Sum(nil)), nil
+		return string(h.Sum(nil)), nil
 	}
 }
 
@@ -68,7 +67,7 @@ func NewMessageHasherSHA256(readLimit int64) MessageHasher {
 		if err != nil && err != io.EOF {
 			return "", err
 		}
-		return hex.EncodeToString(h.Sum(nil)), nil
+		return string(h.Sum(nil)), nil
 	}
 }
 
--- a/message/router/middleware/deduplicator.go
+++ b/message/router/middleware/deduplicator.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"hash/adler32"
 	"io"
+	"math"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


dkotik:
I bumped the coverage to keep Codecov happy.

message/router/middleware/deduplicator.go
```diff
@@ -0,0 +1,237 @@
...
+// or [Deduplicator.Decorator] for a [message.PublisherDecorator].
+//
+// Use [NewMessageHasherAdler32] for fast tagging or
+// [NewMessageHasherSHA256] for minimal collisions.
+//
+// Window specifies the minimum duration of  how long the
+// duplicate tags are remembered for. Real duration can
+// extend up to 50% longer because it depends on the
+// clean up cycle.
+func NewDeduplicator(
```
roblaszczak, line 161: I think that it may be useful to explicitly mention that it's not a solution for at-least-once delivery because it's not persistent, shared between instances etc.

Just to ensure that people won't miss-use it 😉 

WDYT?

message/router/middleware/deduplicator.go
```diff
@@ -0,0 +1,237 @@
...
+	tagger MessageHasher,
+	window time.Duration,
+) *Deduplicator {
+	if tagger == nil {
+		panic("cannot use a <nil> tagger")
```
roblaszczak, line 166: as it's a library, it would be good to return an error here instead - the consumer can handle the error gracefuly (plus we are avoiding all problems related to panics in goroutines etc.)

message/router/middleware/deduplicator.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	if window < time.Millisecond {
+		panic("deduplication window of less than a millisecond is impractical")
+	}
+
+	d := &Deduplicator{
+		tagger: tagger,
+		window: window,
+
+		mu:   &sync.Mutex{},
```
roblaszczak, line 176: I wonder if it won't be worth abstracting it with interface - so you could provide it under in-memory implementation.

It has one nice advantage - someone could use redis/whatever for that and handle more use cases.


WDYT?

message/router/middleware/deduplicator.go
```diff
@@ -0,0 +1,237 @@
...
+		// but this should not be needed for most use cases.
+		d.mu.Unlock()
+		return true, nil
+	}
+	d.tags[tag] = time.Now().Add(d.window)
+	d.mu.Unlock()
+	return false, nil // first time, not a duplicate
+}
+
+// NewDeduplicator returns a new Deduplicator.
```
roblaszczak, line 150: it a bit cosmetic, but what do you think about moving this function to the top of the package - it would enhance readability a lot - usually when you are reading a package to understand it you want to start with constructor with main functionality and by that go further

WDYT?

message/router/middleware/deduplicator.go
```diff
@@ -0,0 +1,237 @@
...
+//
+// Use [NewMessageHasherAdler32] for fast tagging or
+// [NewMessageHasherSHA256] for minimal collisions.
+//
+// Window specifies the minimum duration of  how long the
+// duplicate tags are remembered for. Real duration can
+// extend up to 50% longer because it depends on the
+// clean up cycle.
+func NewDeduplicator(
+	tagger MessageHasher,
```
roblaszczak, line 162: For bigger configs we are usually using structs - so we can add more configuration options later without breaking changes. But it's up to you.

The only forbidden thing is functional option pattern 😈

message/router/middleware/deduplicator.go
```diff
@@ -0,0 +1,237 @@
...
+			continue
+		}
+		notRecent = append(notRecent, m)
+	}
+	return d.Publisher.Publish(topic, notRecent...)
+}
+
+// PublisherDecorator returns a decorator that
+// acknowledges and drops every [message.Message] that
+// was recognized by a [Deduplicator].
```
roblaszczak, line 225: it would be good to explain use case for that

I would also put it a bit up, as it's entrypoint of the public API

dkotik: Adding:

    // The returned decorator provides the same functionality
    // to a [message.Publisher] as [Deduplicator.Middleware]
    // to a [message.Router].

dkotik:
@roblaszczak I resolved all the conversations by doing a bit of a re-write to address the core issue: I added the interface `ExpiringKeyRepository` and flowed everything around that. If somebody implements Redis backend, they can. Perhaps I will later. Requesting re-review.

roblaszczak:
Thanks @dkotik 🍻

PR #425, opened 2024-03-05
Change the autogenerated topic name to conform with Azure limits

### LINKED ISSUE #424: Generated topic names for PubSub tests does not conform with Azure guidelines
When trying to test an incoming azure provider for watermill I have run into an issue that topic names are not conforming with azure guidelines. Azure only allows lowercase letters, digits and dash `-` (if not first or last), while all the topics include `_`, which cause all requests be bounced back with `400 Bad Request`.

unjello:
Azure does only allow letters, digits and dash `-` (if not first, nor last) in the names of Storage Queues.

Closes #424

### DIFF
--- a/pubsub/tests/test_pubsub.go
+++ b/pubsub/tests/test_pubsub.go
@@ -1214,7 +1214,7 @@ func assertConsumerGroupReceivedMessages(
 }
 
 func testTopicName(testID TestID) string {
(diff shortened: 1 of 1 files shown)


m110 (approved):

m110:
Thanks @unjello!

PR #426, opened 2024-03-05
Fix missed topic names non-conforming to Azure policy

unjello:


unjello:
Sorry folks for a mess, but I missed couple changes outside of function :/

m110 (approved):

m110:
Thanks @unjello!

PR #427, opened 2024-03-19
docs: add watermill-rediszset to unofficial libraries

stong1994:
The watermill-rediszset implements a Pub/Sub interface that can be beneficial in certain scenarios, such as scheduled tasks.


roblaszczak:
👍 thx!

PR #444, opened 2024-05-27
Adds feature to use ULID in tests

### LINKED ISSUE #226: Issue with topic and consumer group names in test suite
For an implementation of #126 (PostgreSQL driver with `SKIP LOCKED`), I create a SQL table for each consumer group containing the offsets ready to be consumed. The name for these tables is build by concatenating some prefix, the name of the topic and the name of the consumer group. In some of the test cases in the test suite, UUID are used for both, the topic and the consumer group. Each UUID has a length of 36 characters. Concatenating an UUID for the topic and the consumer group already exceeds the maximum table name length of 63 characters supported by PostgreSQL.

Would it be possible to use a smaller ID for the topic and consumer group names used in the Watermill test suite?

HolyKingCrusader:
Adds feature to use ULID instead of UUID in topic name and consumer group. 

Fixes #226's problem.

### DIFF
--- a/pubsub/tests/test_pubsub.go
+++ b/pubsub/tests/test_pubsub.go
@@ -121,6 +121,9 @@ type Features struct {
 	// NewSubscriberReceivesOldMessages should be set to true if messages are persisted even
 	// if they are already consumed (for example, like in Kafka).
 	NewSubscriberReceivesOldMessages bool
+
+	// UseULID should be set to true if ULID should be used instead of UUID
+	UseULID bool
 }
 
 // RunOnlyFastTests returns true if -short flag was provided -race was not provided.
@@ -151,7 +154,10 @@ func getTestName(testFunc interface{}) string {
 type TestID string
 
 // NewTestID returns a new unique TestID.
-func NewTestID() TestID {
+func NewTestID(useULID bool) TestID {
+	if useULID {
+		return TestID(watermill.NewULID())
+	}
 	return TestID(watermill.NewUUID())
 }
 
@@ -175,7 +181,7 @@ func runTest(
 		if parallel {
 			t.Parallel()
 		}
-		testID := NewTestID()
+		testID := NewTestID(features.UseULID)
 
 		t.Run(string(testID), func(t *testing.T) {
 			tCtx := TestContext{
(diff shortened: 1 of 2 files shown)


pubsub/tests/test_pubsub.go
```diff
@@ -121,6 +121,9 @@ type Features struct {
 	// NewSubscriberReceivesOldMessages should be set to true if messages are persisted even
 	// if they are already consumed (for example, like in Kafka).
 	NewSubscriberReceivesOldMessages bool
+
+	// UseULID should be set to true if ULID should be used instead of UUID
```
m110, line 125:
```suggestion
	// UseULID should be set to true if ULID should be used instead of UUID for generating test IDs.
```

m110:
Thanks for contributing @HolyKingCrusader! :)

pubsub/tests/test_pubsub.go
```diff
@@ -121,6 +121,9 @@ type Features struct {
 	// NewSubscriberReceivesOldMessages should be set to true if messages are persisted even
 	// if they are already consumed (for example, like in Kafka).
 	NewSubscriberReceivesOldMessages bool
+
+	// UseULID should be set to true if ULID should be used instead of UUID for generating test IDs.
+	UseULID bool
```
m110, line 126: @HolyKingCrusader One more idea I had was that we could make it more open for future extensions. How about if there was a `TestIDFunc` (or similar) field which would default to `NewTestID`, but any other could be supplied? (Like `NewTestULID()`)

pubsub/tests/test_pubsub.go
```diff
@@ -154,6 +157,9 @@ type TestID string
 func NewTestID() TestID {
 	return TestID(watermill.NewUUID())
 }
+func NewTestID_ULID() TestID {
```
m110, line 160:
```suggestion
func NewTestULID() TestID {
```
Just to keep the convention of no underscores in func names

pubsub/tests/test_pubsub.go
HolyKingCrusader: @m110 Would it be used anywhere currently or just left for future use?

m110: I guess in scenarios like those mentioned in #226. Similar to what you originally proposed. My point is to make it a bit more open for extension rather than just a UUID/ULID choice. It's not a big deal, though, so let me know your thoughts.

To be clear, I meant having a config field like `TestIDGenerator func() TestID`

HolyKingCrusader: I'll add it, doesn't seem to be too difficult

m110: Hey @HolyKingCrusader, sorry about the delay here 😅 

WIth the field you added to configure the function, the `UseULID` would be redundant now, correct? 

Also, I think the `default` struct tag won't work like this out of the box, you might need to manually check if it's set and default to `NewTestID` otherwise.

PR #461, opened 2024-08-02
Replace protobuf and remove gogo/protobuf

boekkooi-terramate:
This PR replaces the direct dependency on [github.com/golang/protobuf with google.golang.org/protobuf](https://github.com/ThreeDotsLabs/watermill/commit/0e87d46ab9901e75c39e041cfe4bfd8952087b4f) to avoid using a deprecated package. Doing this included adding the generation code to the Makefile to make it more reproducible. 

The PR also removes the [gogo/protobuf](https://github.com/gogo/protobuf) package as it's deprecated and replaced it with google.golang.org/protobuf.


ccoVeille (approved):
LGTM 👍

m110:
Hey @boekkooi-terramate, thanks for contributing. We're aware of the issue with gogo/protobuf, although removing it would be a breaking change, since there might be some code relying on it. How about we mark the marshaler as deprecated and introduce a new one, based on google/protobuf?

PR #465, opened 2024-08-08
Added lock for message.Router.middlewares

SargtLa:
I have faced the race condition here when I was adding a handler with handler middleware while some concurrent process was calling RunHandlers. It is specified that RunHandlers is idempotent and there are locks for handlers, but not for middlewares.

Fixes https://github.com/ThreeDotsLabs/watermill/issues/466


m110:
Thanks @SargtLa!

PR #469, opened 2024-08-20
Requeuer and Message Delay

m110:
* Added the `Requeuer` component.
    * It works as a simpler version of the Forwarder, routing messages from one topic to another (a dynamic one).
    * Can be used to move messages that failed to process back to the original topic, so they don't block other messages.
    * Can be used together with the poison middleware and the `delay` component to delay the forwarding. 
* Added the `delay` package. It contains helpers for setting delay metadata on messages.
    * **Does not do anything by itself.** A Pub/Sub needs to support it explicitly. For now, that's the delayed postgres Pub/Sub implemented in  https://github.com/ThreeDotsLabs/watermill-sql/pull/34
    * Use case 1: publishing a message after a given delay or at given time (see the example).
    * Use case 2: automatically moving messages out of the poison queue to the original topic after a delay (used together with the `Requeuer` component).
* Added the `pq` CLI tool for working with poison queues.


components/requeue/requeue_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	go func() {
+		err := requeue.Run(context.Background())
+		require.NoError(t, err)
+	}()
+
+	router, err := message.NewRouter(message.RouterConfig{}, logger)
+	require.NoError(t, err)
+
+	pq, err := middleware.PoisonQueue(pubSub, "requeue")
```
flowck, line 39: It would be nice to do something like: `pq, err := middleware.PoisonQueue(pubSub, requeue.Topic())`

components/requeue/requeue.go
```diff
@@ -0,0 +1,137 @@
...
+		r.handler,
+	)
+
+	return r, nil
+}
+
+func (r *Requeue) handler(msg *message.Message) error {
+	if r.config.Delay > 0 {
+		time.Sleep(r.config.Delay)
+	}
```
ccoVeille, line 110: I'm always annoyed by such code.

time.Sleep is blocking

Which means if someone defines a 1min delay. The app may not react until the end of this delay, even if the app is shutdown.

I woulld prefer to usage of context 


```suggestion
func (r *Requeue) handler(msg *message.Message) error {
	if r.config.Delay > 0 {
    select {
		    case <- ctx.Done():{
		        return ctx.Err()
		    case <- time.After(r.config.Delay):
	}
```

The main issue might be to find a way to pass the context in this method

m110: thanks 👍

components/requeue/requeue_test.go
m110: Might not be possible anymore since I change the config to include a topic generator. :)

ccoVeille (approved):
I'm glad you liked my feedbacks

components/requeue/requeue.go
```diff
@@ -0,0 +1,139 @@
...
+	config Config,
+	logger watermill.LoggerAdapter,
+) (*Requeue, error) {
+	config.setDefaults()
+	err := config.validate()
+	if err != nil {
+		return nil, fmt.Errorf("invalid config: %w", err)
+	}
+
+  router := config.Router
```
ccoVeille, line 82: I think this line is miss aligned because of my suggestion.

Using gofmt would solve the issue

m110: This is still a rough draft / Poc 🙃

ccoVeille: But I'm a crazy nitpicking person 🙋‍♂️😂🤦‍♂️

_examples/real-world-examples/delayed-messages/main.go
```diff
@@ -0,0 +1,128 @@
...
+	_ "github.com/lib/pq"
+
+	"github.com/ThreeDotsLabs/watermill"
+	"github.com/ThreeDotsLabs/watermill-sql/v3/pkg/sql"
+	"github.com/ThreeDotsLabs/watermill/components/cqrs"
+	"github.com/ThreeDotsLabs/watermill/components/delay"
+	"github.com/ThreeDotsLabs/watermill/message"
+)
+
+func main() {
```
roblaszczak, line 20: I wonder if we could not have a bit more real-life scenario here, like sending reminder when payment was not sent after X time or whatever?

or something what should happen after X after order was placed?

_examples/real-world-examples/delayed-requeue/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+services:
+  server:
+    image: golang:1.23
+      #restart: unless-stopped
+    volumes:
+      - .:/app
+      - $GOPATH/pkg/mod:/go/pkg/mod
+    working_dir: /app
+      #command: go run main.go
```
roblaszczak, line 9: I guess it's because it's WIP?

components/delay/delay.go
```diff
@@ -0,0 +1,65 @@
...
+	"time"
+
+	"github.com/ThreeDotsLabs/watermill/message"
+)
+
+// Delay represents a message's delay.
+// It can be either a delay until a specific time or a delay for a specific duration.
+// The zero value of Delay is a zero delay.
+//
+// IMPORTANT: Delay doesn't work with all Pub/Subs! Using it won't have any effect on Pub/Subs that don't support it.
```
roblaszczak, line 14: I wonder if we can somehow ensure that someone is not using Pub/Sub that doesn't support it 🤔 

as far as I understand, now you can use whatever pubsub and if it's not compatible it will "silently not work"?

I can imagine that if someone can accedantly inject wrong one, it won't be visible

it would be nice if we could do it compile time, or at least return some error so it won't be totally silent

roblaszczak, line 14: I woner if we don't want to put list of supported Pub/Subs for that somewhere? it may be not obvious to find

components/delay/publisher.go
```diff
@@ -0,0 +1,50 @@
+package delay
+
+import "github.com/ThreeDotsLabs/watermill/message"
+
+// PublisherConfig is a configuration for the delay publisher.
+type PublisherConfig struct {
+	DefaultDelay Delay
```
roblaszczak, line 7: maybe closure that gets the topic/message/struct with params?

tools/pq/backend/postgres.go
```diff
@@ -0,0 +1,100 @@
...
+	Payload  string `db:"payload"`
+	Metadata string `db:"metadata"`
+}
+
+type PostgresBackend struct {
+	db     *sqlx.DB
+	config cli.BackendConfig
+}
+
+func NewPostgresBackend(ctx context.Context, config cli.BackendConfig) (*PostgresBackend, error) {
```
roblaszczak, line 29: food for thought: how to move it to Postgres adapter potentially 🤔

_examples/real-world-examples/delayed-requeue/docker-compose.yml
m110: Because of go.mod replace, removed

components/delay/delay.go
m110: As discussed, this is not exactly possible, but I added the opposite: b7fb98bf9b08d4b31627eee3400368bfd262b2d4

m110: Added also `AllowNoDelay` to the delayed subscriber's config.

m110: Added docs eaa5ff93814041fc6c37339591c394561d37c79f

tools/pq/backend/postgres.go
m110: As discussed, let's keep it as is for now, and see if there's a better abstraction in the future.

tools/pq/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Install
+
+```bash
+go install github.com/ThreeDotsLabs/watermill/tools/pq@latest
+```
+
+## Usage
+
+Set the `DATABASE_URL` environment variable to your PostgreSQL connection string.
```
roblaszczak, line 15: it would be nice to provide example one - they are 10 different formats and it's useful to have it

tools/pq/README.md
```diff
@@ -0,0 +1,32 @@
+# pq
```
roblaszczak, line 1: probably it would be nice to link here from poison queue godoc?

m110: Which one, though? The SQL one?

tools/pq/README.md
m110: cc91ad38bf2c9219e20042be5ea46335c02e1c06

tools/pq/README.md
m110: Will add more docs in a follow-up PR.

Issue #464, opened 2024-08-05
replace original context in pub/sub GoChannel (SendToSubscriber)


bi0dread:
Hi
you have replaced original msg context with msgToSend.SetContext(ctx) and we lost the original context!

```
func (s *subscriber) sendMessageToSubscriber(msg *message.Message, logFields watermill.LogFields) {
	s.sending.Lock()
	defer s.sending.Unlock()

	ctx, cancelCtx := context.WithCancel(s.ctx)
	defer cancelCtx()

SendToSubscriber:
	for {
		// copy the message to prevent ack/nack propagation to other consumers
		// also allows to make retries on a fresh copy of the original message
		msgToSend := msg.Copy()
		msgToSend.SetContext(ctx)

		s.logger.Trace("Sending msg to subscriber", logFields)

		if s.closed {
			s.logger.Info("Pub/Sub closed, discarding msg", logFields)
			return
		}

		select {
		case s.outputChannel <- msgToSend:
			s.logger.Trace("Sent message to subscriber", logFields)
		case <-s.closing:
			s.logger.Trace("Closing, message discarded", logFields)
			return
		}

		select {
		case <-msgToSend.Acked():
			s.logger.Trace("Message acked", logFields)
			return
		case <-msgToSend.Nacked():
			s.logger.Trace("Nack received, resending message", logFields)
			continue SendToSubscriber
		case <-s.closing:
			s.logger.Trace("Closing, message discarded", logFields)
			return
		}
	}
}
```

m110:
Hey @bi0dread. This is to maintain the same behavior as other Pubs/Subs. Other backends transfer the message over the network, so they can't keep the context. I guess we could make it an option in the config to keep the context, though, as a special case.

yashb042:
@m110  This issue is what I am also facing. At least in the local pub-sub case, the context must be propagated (even though it doesn't match the behaviour of other pub-subs. 

Let me know if you want me to add this feature and push.

m110:
Hey @yashb042. I think it makes sense, just please keep it as a config option with proper disclaimer that it doesn't match the generic behavior. :)

yashb042:
Okay, picking this up

yashb042:
https://github.com/ThreeDotsLabs/watermill/pull/487/files

Adding test cases soon
Please let me know if the variable addition is okay.
Have not added extra variables in Message struct, because the Message struct is used by every publisher-subscriber

yashb042:
[...]

PR #487, opened 2024-09-03
Add context propogation to subscribers for local pub-sub

yashb042:
Solves Issue[https://github.com/ThreeDotsLabs/watermill/issues/464](url)


pubsub/gochannel/pubsub.go
```diff
@@ -340,11 +349,18 @@ func (s *subscriber) Close() {
...
+func (s *subscriber) sendMessageToSubscriber(msg *message.Message, logFields watermill.LogFields, preserveContext bool) {
 	s.sending.Lock()
 	defer s.sending.Unlock()
 
-	ctx, cancelCtx := context.WithCancel(s.ctx)
+	var ctx context.Context
+	var cancelCtx context.CancelFunc
+
+	if preserveContext {
+		ctx, cancelCtx = context.WithCancel(msg.Context())
```
m110, line 360: Can you describe what's the use case here? 🤔 It seems to be a bit more than just preserving context among messages.

m110:
Thank you @yashb042! This looks solid to me, just a couple of minor comments.

pubsub/gochannel/pubsub.go
```diff
@@ -340,11 +349,18 @@ func (s *subscriber) Close() {
 	close(s.outputChannel)
 }
 
-func (s *subscriber) sendMessageToSubscriber(msg *message.Message, logFields watermill.LogFields) {
+func (s *subscriber) sendMessageToSubscriber(msg *message.Message, logFields watermill.LogFields, preserveContext bool) {
```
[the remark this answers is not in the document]
yashb042: Moved preserveContext inside the subscriber.

pubsub/gochannel/pubsub.go
yashb042: Excellent find, removed the WithCancel when preserving context.

WithCancel makes sense when we are using the subscriber context, and done is called on the subscriber context itself. 

A quick question though - Even if the subscriber main context is cancelled, that time the messages will still be processed further.

yashb042:
Also, @m110 

Update 2 -
Couldn't test the func - TestMessageCtx in test_pubsub.go
It's mentioned that "ExactlyOnceDelivery test is not supported yet"

How would we proceed testing this func -  **TestMessageCtx**

pubsub/gochannel/pubsub.go
```diff
@@ -344,8 +356,16 @@ func (s *subscriber) sendMessageToSubscriber(msg *message.Message, logFields wat
...
+	var ctx context.Context
+
+	//This is getting the context from the message, not the subscriber
+	if s.preserveContext {
+		ctx = msg.Context()
+	} else {
+		var cancelCtx context.CancelFunc
+		ctx, cancelCtx = context.WithCancel(s.ctx)
+		defer cancelCtx()
+	}
```
ccoVeille, line 368: I would like to suggest this


```suggestion
	ctx := msg.Context()

	//This is getting the context from the message, not the subscriber
	if !s.preserveContext {
		var cancelCtx context.CancelFunc
		ctx, cancelCtx = context.WithCancel(s.ctx)
		defer cancelCtx()
	}
```

pubsub/tests/test_asserts.go
```diff
@@ -92,3 +93,19 @@ func AssertMessagesMetadata(t *testing.T, key string, expectedValues map[string]
...
+	for _, msg := range received {
+		expectedValue := expectedValues[msg.UUID].Value(contextKey(contextKeyString)).(string)
+		actualValue := msg.Context().Value(contextKeyString)
+		if !assert.Equal(t, expectedValue, actualValue) {
+			ok = false
+		}
+	}
+
+	return ok
+}
```
ccoVeille, line 111: You don't need to return a boolean, calling t.Error is enough to invalidate the test

also your code is not using the returned value when you call it

So this would be better I think

```suggestion
func AssertAllMessagesHaveSameContext(t *testing.T, contextKeyString string, expectedValues map[string]context.Context, received []*message.Message) {
	assert.Len(t, received, len(expectedValues))

	for _, msg := range received {
		expectedValue := expectedValues[msg.UUID].Value(contextKey(contextKeyString)).(string)
		actualValue := msg.Context().Value(contextKeyString)
		assert.Equal(t, expectedValue, actualValue)
	}
}
```

ccoVeille (changes requested):

pubsub/tests/test_asserts.go
yashb042: Agreed, changed
Thanks for the suggestion

pubsub/gochannel/pubsub.go
yashb042: Agreed, changed

A quick question though :-

When we are using the subscriber context, even if the subscriber main context is cancelled, that time the messages will still be processed further.
So why are we even using cancelCtx here.

ccoVeille: To be honest, I was surprised to see an option to change it, for me the context should be always the one of the message

The lib is in charge to pass the context to the message, so if the context is cancelled, the message one is too, no ?

I mean couldn't we simply fix it, and not adding the preserveContext boolean ?

yashb042: I agree, however @m110 has a point here - 
https://github.com/ThreeDotsLabs/watermill/issues/464#issuecomment-2283885824

Nonetheless, according to your statement - 
`The lib is in charge to pass the context to the message, so if the context is cancelled, the message one is too, no ?
`

I don't see that happening in the code though, even if the main context is cancelled, the messages are passed to the subscriber nonetheless

pubsub/gochannel/pubsub.go
m110: Hey @yashb042 after considering it, I think your original idea made more sense: The message's context should also be wrapped with the cancel, otherwise closing the subscriber wouldn't close the message's context. Sorry about the confusion!

ccoVeille:
Before reviewing, I will wait for you to reply to this message, because I feel like the code might change again https://github.com/ThreeDotsLabs/watermill/pull/487#discussion_r1749848993

pubsub/gochannel/pubsub.go
ccoVeille: Yes, they are, but the context of the message would be canceled I think/hope

Anyway this comment might change everything https://github.com/ThreeDotsLabs/watermill/pull/487#discussion_r1749848993

So, let's wait

pubsub/gochannel/pubsub.go
yashb042: @m110 Could you please tell me what is the point of keeping cancelCtx here, for both the cases
As I see it, the messages are still being passed to the subscriber irrespective of whether the context was cancelled or not.

m110: @yashb042 It's because the context is passed to the message and is likely to be used in the command handler. For example, when executing a database query, you would pass `msg.Context()`. If the subscriber closes, the context gets canceled, which also cancels the database query. 

Let me know if this makes sense?

ccoVeille: It does make sense

But let me ask something.

Is there a case where there is a need for the context not to be the one of the message ?

I'm asking because the previous code was creating a context. And now we add an option to preserve it.

But why do we need such option? Why not simply fix the code to preserve the context.

I mean unless I'm wrong the code has a bug, and we could simply fix it.

m110: Most Watermill Pub/Subs work over the network. You publish a message, it gets marshalled, then it's delivered to the subscriber. But obviously, the context wouldn't survive the marshaling, so you can't rely on it.

The GoChannel Pub/Sub works in memory, so it is possible to pass the context. I suppose it could be useful for passing some arbitrary data instead of marshaling it to `[]byte`, but it behaves by design the same as other Pub/Subs. Because you should be able to replace one Pub/Sub implementation with another with similar results. For example, it's handy to use GoChannel for tests, but you would use something different for production.

A scenario where passing the context may be confusing:
* An HTTP handler publishes a message passing the request's context.
* The message is delivered to a subscriber, which appends some data to the database, using the message's context.
* Meanwhile, the HTTP request is done, and the context is canceled. If you preserved the context in the message, it would also cancel the database query, which is not what you want.

In general, the publishing and subscribing sides should be considered as something separate, working concurrently. Preserving the context can make things confusing.

If you really know what you're doing, I guess it's fine to change this behavior, but you should explicitly flip the config flag then and proceed with caution (e.g., use `context.WithoutCancel` when passing it to the `Publish`).

ccoVeille: Thank you for your detailed reply!

yashb042: > @yashb042 It's because the context is passed to the message and is likely to be used in the command handler. For example, when executing a database query, you would pass `msg.Context()`. If the subscriber closes, the context gets canceled, which also cancels the database query.
> 
> Let me know if this makes sense?

Agreed, and the database case makes sense as well.
Though, is it there for all the cases, because as I see it, context usually is considered whenever either you do a network call (in database case you mentioned) or you call a library which has put up an exclusive check on the context.Done() field.

For our local pub-sub, I don't see any of these happening that's why a bit confused.

ccoVeille (approved):

ccoVeille:
I find back this old PR.

Is there anything blocking or to be discussed that prevent it to be merged?

roblaszczak (approved):

PR #502, opened 2024-10-14
Use errors.Is for error comparison instead of direct equality checks

ytnsym:
This PR refactors error handling by replacing direct comparisons of errors (e.g., err == someError) with the use of errors.Is.

### DIFF
--- a/message/router/middleware/poison_test.go
+++ b/message/router/middleware/poison_test.go
@@ -2,21 +2,20 @@ package middleware_test
 
 import (
 	"context"
+	stdErrors "errors"
 	"testing"
 	"time"
 
-	"github.com/ThreeDotsLabs/watermill"
-	"github.com/ThreeDotsLabs/watermill/message/subscriber"
-	"github.com/ThreeDotsLabs/watermill/pubsub/gochannel"
-
 	"github.com/hashicorp/go-multierror"
-
-	"github.com/ThreeDotsLabs/watermill/message"
-
-	"github.com/ThreeDotsLabs/watermill/message/router/middleware"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/ThreeDotsLabs/watermill"
+	"github.com/ThreeDotsLabs/watermill/message"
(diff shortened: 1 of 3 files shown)


message/router/middleware/poison_test.go
```diff
@@ -2,21 +2,20 @@ package middleware_test
...
-	"github.com/ThreeDotsLabs/watermill"
-	"github.com/ThreeDotsLabs/watermill/message/subscriber"
-	"github.com/ThreeDotsLabs/watermill/pubsub/gochannel"
-
 	"github.com/hashicorp/go-multierror"
-
-	"github.com/ThreeDotsLabs/watermill/message"
-
-	"github.com/ThreeDotsLabs/watermill/message/router/middleware"
 	"github.com/pkg/errors"
```
ccoVeille, line 10: I dont see a need for this deprecated one

Remove it, so you can use errors directly without an alias

The call to errors.Wrap can be replaced by fmt.Errorf("%w ...")

errors.New can be used directly from stdlib

ccoVeille:
Also, maybe it's time to enable some golangci-lint linters

https://golangci-lint.run/usage/linters/

err113 is exacly the linter you need for catching what you are fixing.

errorlint is a good addition

Of course, this could be considered separately from this PR

message/router/middleware/poison_test.go
ytnsym: It looks like errors.Cause is being used in the current implementation, which makes it a bit more challenging to simply remove the alias.

ccoVeille: From what I had seen before raising my point, errors.Cause is used only in the code, not in the test file I quoted.

But, I have missed something. And I was wrong

There is this indeed.

```go
assert.Equal(t, errFailed, errors.Cause(multierr.WrappedErrors()[1]))
```

this could be rewritten to use maybe

```go
assert.ErrorIs(t, multierr, errFailed)
```

ccoVeille (approved):
Feel free to resolve directly my comment, or fix if you want to try.

But for me my comment was a detail, I'm fine with the code.

And thank you for opening the PR

m110:
Thank you @ytnsym!

PR #510, opened 2024-10-27
Slog level mapping

roblaszczak:


m110 (approved):

docs/hugo_stats.json
```diff
@@ -317,7 +317,6 @@
       "debugging-pubsub-tests",
       "deduplicator",
       "delay-metadata",
-      "delay-on-error",
```
ccoVeille, line 320: Why this change?

roblaszczak: this is some garbage generated by hugo, I gitignored it

PR #514, opened 2024-11-03
Example SSE: lock the versions of used Vue package so the example runs.

thpk:
One of the Server Sent Events examples didn't work anymore, probably due to newer Vue versions existing.

By locking the versions the example runs.


m110:
Thank you @thpk! 🫡🎖️

PR #515, opened 2024-11-03
fix: first added middleware should be executed first

Azzahid:
Reverse the middleware execution order.
router.go:210 and router.go:197 use to append middleware instead of prepend it.

Reversing the loop so it will execute the middleware that got added first.


m110:
The current order is correct. You have to be careful with the order, depending on which middleware you use. Some of them, like `Retry`, execute the entire chain of handlers and then react to the error, which can be counterintuitive.

PR #517, opened 2024-11-21
Restore message context in Timeout middleware

### LINKED ISSUE #467: Timeout middleware: context is not restored
If you take a look at the Timeout middleware: 

```
func Timeout(timeout time.Duration) func(message.HandlerFunc) message.HandlerFunc {
	return func(h message.HandlerFunc) message.HandlerFunc {
		return func(msg *message.Message) ([]*message.Message, error) {
			ctx, cancel := context.WithTimeout(msg.Context(), timeout)
			defer func() {
				cancel()
			}()

			msg.SetContext(ctx)
			return h(msg)
		}
	}
}
```

It is visible that the original context is never set to the original value if a timeout occurs. However the Retry middleware uses the message's context to ensure that the MaxElapsedTime is [supported](https://github.com/ThreeDotsLabs/watermill/blob/33bd35f3cde5b14b5d1ffdcc53e21b9e8663b09e/message/router/middleware/retry.go#L54-L58):

```
                 if r.MaxElapsedTime > 0 {
			var cancel func()
			ctx, cancel = context.WithTimeout(ctx, r.MaxElapsedTime)
			defer cancel()
		}
```

This means that if I have this chain Retry(Timeout(handler)), and a timeout occurs then the Retry will not be effective.

I would recommend to save the original context in Timeout and restore that in the defer function(). Or is this the expected behavior?

zanvd:
This resolves #467.

It simply stores the original message context and restores it in the `defer`.

### DIFF
--- a/message/router/middleware/timeout.go
+++ b/message/router/middleware/timeout.go
@@ -9,11 +9,16 @@ import (
 
 // Timeout makes the handler cancel the incoming message's context after a specified time.
 // Any timeout-sensitive functionality of the handler should listen on msg.Context().Done() to know when to fail.
+// Before exiting this middleware, message's original context is restored to not disrupt the logic of possible
+// enveloping handlers (e.g. Retry) that also rely on message context.
 func Timeout(timeout time.Duration) func(message.HandlerFunc) message.HandlerFunc {
 	return func(h message.HandlerFunc) message.HandlerFunc {
 		return func(msg *message.Message) ([]*message.Message, error) {
+			orgCtx := msg.Context()
+
 			ctx, cancel := context.WithTimeout(msg.Context(), timeout)
 			defer func() {
+				msg.SetContext(orgCtx)
 				cancel()
 			}()
 
--- a/message/router/middleware/timeout_test.go
+++ b/message/router/middleware/timeout_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/ThreeDotsLabs/watermill/message"
@@ -25,6 +26,9 @@ func TestTimeout(t *testing.T) {
 		}
 	})
 
-	_, err := h(message.NewMessage("any-uuid", nil))
+	msg := message.NewMessage("any-uuid", nil)
+
+	_, err := h(msg)
 	require.NoError(t, err)
+	assert.Nil(t, msg.Context().Err())
 }


message/router/middleware/timeout.go
```diff
@@ -9,11 +9,16 @@ import (
...
 // Timeout makes the handler cancel the incoming message's context after a specified time.
 // Any timeout-sensitive functionality of the handler should listen on msg.Context().Done() to know when to fail.
+// Before exiting this middleware, message's original context is restored to not disrupt the logic of possible
+// enveloping handlers (e.g. Retry) that also rely on message context.
 func Timeout(timeout time.Duration) func(message.HandlerFunc) message.HandlerFunc {
 	return func(h message.HandlerFunc) message.HandlerFunc {
 		return func(msg *message.Message) ([]*message.Message, error) {
+			orgCtx := msg.Context()
+
 			ctx, cancel := context.WithTimeout(msg.Context(), timeout)
```
ccoVeille, line 19:
```suggestion
			ctx, cancel := context.WithTimeout(orgCtx, timeout)
```

ccoVeille (approved):
A minor remark

ccoVeille (approved):

message/router/middleware/timeout.go
```diff
@@ -9,11 +9,16 @@ import (
...
+// enveloping handlers (e.g. Retry) that also rely on message context.
 func Timeout(timeout time.Duration) func(message.HandlerFunc) message.HandlerFunc {
 	return func(h message.HandlerFunc) message.HandlerFunc {
 		return func(msg *message.Message) ([]*message.Message, error) {
-			ctx, cancel := context.WithTimeout(msg.Context(), timeout)
+			orgCtx := msg.Context()
+
+			ctx, cancel := context.WithTimeout(orgCtx, timeout)
 			defer func() {
+				msg.SetContext(orgCtx)
```
m110, line 21: @zanvd how about we use `context.WithoutCancel(ctx)` here? This way the cancellation won't be propagated while the values stored in the context can be. Would that also solve the root issue? 🤔

m110: Ah, you're right.

My point is that restoring the context will also remove all context values. If someone uses middleware that uses the context after the handler is executed, it will be lost this way.

ccoVeille: Here is the way then


```go
orgCtx := msg.Context() 
ctx, cancel := context.WithTimeout(orgCtx, timeout)

defer func() {
   cancel()
   ctx, cancel2 := context.WithCancelWithCause(context.WithoutCancel(ctx))
   msg.SetContext(ctx)
   go func() {
        <- ctx.Done()
        cancel2(ctx.Err)
   }
}()
```

But I find it ugly

zanvd: I'd have to test it out, but this is more or less why I was asking on Discord about usage of context within `Message` - makes it sort of a global object that anyone can modify. Since the blueprint for this was the net/http package, I'd follow its [logic](https://cs.opensource.google/go/go/+/refs/tags/go1.23.3:src/net/http/request.go;l=332) and not even allow changing of the `Message`'s context. Not something to solve in this PR, though.

What would be an example of values stored in the context that we should be vary of? Perhaps copying them to the original context might be a temporary solution.

ccoVeille:
Is there anything blocking or to be discussed that prevents to consider reviwing the PR and merge it?

roblaszczak:
After discussing with @m110 we decided that it would be the best to restore context Retry middleware: https://github.com/ThreeDotsLabs/watermill/pull/598

Retry is a bit better place, because someone can implement a custom `Timeout` middleware.

It's disabled by default to avoid breaking changes (it's a terrible idea, but in theory someone can depend on passing values between retries).

PR #460, opened 2024-07-31
Use errors and fmt package to join and wrap errors

boekkooi-terramate:
Good day!

First of thank you for maintaining and providing the watermill package!

I opened this PR because I noticed that the `hashicorp/go-multierror` and `github.com/pkg/errors` where added as an indirect dependency when using watermill.
As these dependency have native alternatives I created this PR.

Since golang 1.20 the go `errors` package supports `errors.Join` which replaces the need for the `hashicorp/go-multierror` package which was providing the same functionality. 
To wrap errors `fmt.Errorf` with the `%w` format specifier (introduced in golang 1.13) is now used instead of `errors.Wrap` which was provided by the `github.com/pkg/errors` package.

Please let me know what you think and if any changes are needed.
Have a great day and thank you for reviewing this PR!

Cheers,
Warnar

Issue #447, opened 2024-06-03
Removing usage of https://github.com/hashicorp/go-multierror


elordis:
@roblaszczak, @m110
Hello, team.
Our security guys recently noticed that Watermill uses https://github.com/hashicorp/go-multierror package, which uses MPL-2.0 License, which is a bit of a pain for us. After a quick look at the watermill code I've noticed that it would be trivial to replace go-multierror usage with standard library implementation. The only downside that I see is that `errors.Join` appeared in Go 1.20, but current Watermill is already at 1.21 and I didn't notice any version compatibility guarantees in documentation.

Are there any plans for such a change? Are you willing to accept a Pull request if we do it ourselves?

dkotik:
I believe this is important to get done. It seems that #460 was trying to do too many changes at once. @elordis submit perhaps just the errors.Join patch? May get to it myself later this year.

roblaszczak:
Solved by https://github.com/ThreeDotsLabs/watermill/pull/550

PR #550, opened 2025-04-01
Remove hashicorp/go-multierror

co60ca:
### Motivation / Background

hashicorp/go-multierror has a license that is problematic to our projects. There has also been interested to remove it from other parties. See https://github.com/ThreeDotsLabs/watermill/issues/447

### Detail

This rebuilds part of the solution from https://github.com/ThreeDotsLabs/watermill/pull/460 , which arguably failed to land due to trying to do too much (from what I can tell)

We could change `stdErrors` -> `errors` and then change `pkg/errors` to `pkgErrors`. Or attempt to also remove `pkg/errors` but I'm attempting to keep this specific PR small. Once this change is made it is extremely easy using refactoring tools to change the name of these packages. Removing pkgErrors could be a small PR on its own. Replacing errors.Wrapf with fmt.Errorf is a tedious operation.

Note, its probably better performance wise to do `var err []errors` and then call `errors.Join` right near the end, but it might be a bit of pre-mature optimization considering how often we expect that to happen. `errors.Join` returns a nil of the input is empty.

Also note, I did not remove it from the examples, because it requires a new version number, which can be done in bulk afterwards.

### Checklist

The resources of our team are limited. **There are a couple of things that you can do to help us merge your PR faster**:

- [X] I wrote tests for the changes.
_I modified the minimum tests respecting their original intention_
- [X] All tests are passing.
  - If you are testing a Pub/Sub, you can start Docker with `make up`.
  - You can start with `make test_short` for a quick check.
  - If you want to run all tests, use `make test`.
- [ ] Code has no breaking changes.
_Its possible this code could break someones tests if expects a multierror type specifically. People should generally expect typed errors from packages that are exported from that package using `errors.Is` or `errors.As` this would still work if they are comparing using strings._
~~- [ ] _(If applicable)_ documentation on [watermill.io](https://watermill.io/) is updated.~~
~~- Documentation is built in the [github.com/ThreeDotsLabs/watermill/docs](https://github.com/ThreeDotsLabs/watermill/tree/master/docs).~~
~~- You can find development instructions in the [DEVELOP.md](https://github.com/ThreeDotsLabs/watermill/tree/master/docs/DEVELOP.md).~~


roblaszczak (approved):
Thanks!

co60ca:
@roblaszczak looks like we need you to re-approve the merge in order to get CI to run.

roblaszczak:
Done!

PR #561, opened 2025-05-05
add memory clean up on gochannel subscribe topics when they no longer have any more subscribers

dkotik:
This allows publishing and subscribing to narrowly focused topics that include random data like UUIDs in topic name. Without this change, memory usage will grow indefinitely in a long-running service as the map grows larger and larger with keys pointing to empty slices.

The change is small, but it allows using Watermill Gochannel like NATS - having a topic per HTTP request, for example.


pubsub/gochannel/pubsub.go
```diff
@@ -260,6 +260,16 @@ func (g *GoChannel) removeSubscriber(topic string, toRemove *subscriber) {
...
 			removed = true
+
+			if len(g.subscribers[topic]) == 0 && !g.config.Persistent {
+				// Free up the memory taken by a topic which no longer has subscribers.
+				// This operation allows publishing and subscribing to narrowly
+				// focused topics that include random data like UUIDs in topic name.
+				//
+				// Without this operation, memory usage will grow indefinitely in a long-running service
+				// as the map grows larger and larger with keys pointing to empty slices.
+				delete(g.subscribers, topic)
```
m110, line 271:
```suggestion
				delete(g.subscribers, topic)
				g.subscribersByTopicLock.Delete(topic)
```

PR #564, opened 2025-05-11
Update watermill to github.com/cenkalti/backoff/v4

StevenACoffman:
This is an incremental update to use cenkalti/backoff/v4 rather than cenkalti/backoff/v3

https://github.com/cenkalti/backoff/releases/tag/v3.2.2 was released in 2020
https://github.com/cenkalti/backoff/releases/tag/v4.3.0 and 4.3.0 was released in 2024

https://github.com/cenkalti/backoff/compare/v0.17.72...v0.17.73


StevenACoffman:
@roblaszczak @m110 I would like your thoughts on adapting watermill's bespoke Retry or using the backoff v5 Retry so we can continue to track security and performance improvements from upstream.

Issue #382, opened 2023-09-04
Blocking in GoChannel `sendMessageToSubscriber`


qiulin:
My use case is reading kafka message using segment-io kafka driver and publishing to watermill using GoChannel. One day I got a kafka topic blocking. After read watermill code, I found it blocked at `case <-msgToSend.Acked():`, but logs show `msg.Ack()` is ok, and "Message acked" in `handler.handleMessage` was printed.

``` go
func (s *subscriber) sendMessageToSubscriber(msg *message.Message, logFields watermill.LogFields) {
	s.sending.Lock()
	defer s.sending.Unlock()

	ctx, cancelCtx := context.WithCancel(s.ctx)
	defer cancelCtx()

SendToSubscriber:
	for {
		// copy the message to prevent ack/nack propagation to other consumers
		// also allows to make retries on a fresh copy of the original message
		msgToSend := msg.Copy()
		msgToSend.SetContext(ctx)

		s.logger.Trace("Sending msg to subscriber", logFields)

		if s.closed {
			s.logger.Info("Pub/Sub closed, discarding msg", logFields)
			return
		}

		select {
		case s.outputChannel <- msgToSend:
			s.logger.Trace("Sent message to subscriber", logFields)
		case <-s.closing:
			s.logger.Trace("Closing, message discarded", logFields)
			return
		}

		select {
		case <-msgToSend.Acked():
			s.logger.Trace("Message acked", logFields)
			return
		case <-msgToSend.Nacked():
			s.logger.Trace("Nack received, resending message", logFields)
			continue SendToSubscriber
		case <-s.closing:
			s.logger.Trace("Closing, message discarded", logFields)
			return
		}
	}
}

```

m110:
Addressed in https://github.com/ThreeDotsLabs/watermill/pull/589

PR #589, opened 2025-07-30
Refactored sending lock of subscriber

stephanvanzwienen:
### Motivation / Background

Refactored sending lock of subscriber to the SendToSubscriber loop to be as close as possible to the sending part of the logic. This results in the output buffer actually being utilized and fixes blocking behavior mentioned in issue #275 and #382.
### Detail

In a project we came to a point where we wanted to rely on "flushing" behavior noted in the docs [here](https://watermill.io/docs/pub-sub/#close). Which we expected to work as go channels can still be read from even after calling close as per [doc](https://pkg.go.dev/builtin#close:~:text=It%20should%20be%20executed%20only%20by%20the%20sender%2C%20never%20the%20receiver%2C%20and%20has%20the%20effect%20of%20shutting%20down%20the%20channel%20after%20the%20last%20sent%20value%20is%20received)

But to our surprise in testing where we set up a PubSub and published an equal amount of messages as the buffer and then calling close on the pubsub. We only received a fraction of the messages in our subscriber handler.  Note that we made sure to only start reading from the channel after a few seconds, which should be plenty amount of time, to give the publisher a change to fill up the buffer. We noticed that the buffer didn't contain the expected published messages. But we did notice the messages being present int the `persistedMessages` map when making usage of the persist config. So timing couldn't being the issue.
I tracked down the culprit of this (for us) unexpected behavior to the `s.sending.Lock()` being in a place where it introduces a deadlock. 

When Publishing() multiple messages each message result in a go routine which loops over all the subscriber of the targeted topic.  For each subscriber a new go routine is created which finally send the actual message through `subscriber.sendMessageToSubscriber(message, logFields)`.

However, ALL go routines share the same "sending lock". Thus, if multiple messages are sent concurrently 1 go routine will keep this lock hostage ("deadlock") until it returns from this `sendMessageToSubscriber` func. This is due the unlock being done on defer AND because of the last switch case where the func only returns if:
1. the message is acked
2. the message is nacked
3. s.closing is received
And this switch case could take a while to actually trigger depending on when a dev calls one of these 3 options. 
This part of the logic has no further reason (as far as I understand) to keep this sending lock. 

Now by moving the lock /unlock to the "sending" part of the logic we avoid this deadlock.  And circles back to our issue where now the buffer is actually utilized and flushing can be done. While also fixing #275 and #382.
 
N/A

### Checklist

The resources of our team are limited. **There are a couple of things that you can do to help us merge your PR faster**:

- [X] I wrote tests for the changes.
- [X] All tests are passing.
  - If you are testing a Pub/Sub, you can start Docker with `make up`.
  - You can start with `make test_short` for a quick check.
  - If you want to run all tests, use `make test`.
- [X] Code has no breaking changes.
- [ ] _(If applicable)_ documentation on [watermill.io](https://watermill.io/) is updated.
  - Documentation is built in the [github.com/ThreeDotsLabs/watermill/docs](https://github.com/ThreeDotsLabs/watermill/tree/master/docs).
  - You can find development instructions in the [DEVELOP.md](https://github.com/ThreeDotsLabs/watermill/tree/master/docs/DEVELOP.md).


m110 (approved):

m110:
Great investigation @stephanvanzwienen, thank you for contributing!

stephanvanzwienen:
@m110 when can we expect the next release?

PR #667, opened 2026-02-03
Added support for custom JSON configuration in the forwarder component

falboe:
This pull request introduces the ability to inject a custom JSON marshal/unmarshal implementation into the Forwarder component.

This makes the component more flexible and allows consumers to use alternative JSON libraries (e.g. Sonic) instead of being tightly coupled to the standard encoding/json package.


m110:
Thanks for contributing @falboe! I extended the interface a bit to be consistent with how other configs work. :)

roblaszczak (approved):

PR #670, opened 2026-03-08
feat(retry): add OnRetriesExhausted hook for cleanup actions

### LINKED ISSUE #545: Is there any way to write clean up hook after max retries?
This is just simple issue I had in mind. In my usecases, every time I retry, I change the entry in my databases to indicate how many retry in there. This case, is could be solved by using ``OnRetryHook`` on the Retry Middleware.

In my usecases, when the retry is at max limit, I want to do an action to do something (i.e notify me to Discord or anythings). What I had in mind was this....

```go
type Retry struct {
       // All the atributes goes in here....
 
	// OnRetryHook is an optional function that will be executed on each retry attempt.
	// The number of the current retry is passed as retryNum,
	OnRetryHook func(retryNum int, delay time.Duration)

        OnCleanUpHook func(msg *message.Message) // something like this
}
```

Is there any better way to do this or the package has built-in function to solve this?
Thanks in advance

Br1an67:
Fixes #545

### Motivation / Background

This PR adds an `OnRetriesExhausted` hook to the Retry middleware, allowing users to execute cleanup actions when all retry attempts are exhausted. This addresses the need for performing actions like sending notifications (e.g., to Discord) or updating database records when a message handler consistently fails after the maximum number of retries.

### Detail

Added a new optional field `OnRetriesExhausted` to the `Retry` struct that accepts a function with signature `func(err error, retryNum int)`. This hook is called when:
- The maximum number of retries is reached
- The handler still returns an error

The hook is NOT called when:
- The handler succeeds
- `ShouldRetry` returns false (which results in a permanent error)

Changes:
- Added `OnRetriesExhausted func(err error, retryNum int)` field to the `Retry` struct
- Call the hook after backoff.Retry completes with an error (excluding permanent errors)
- Added tests to verify the hook is called when retries are exhausted and not called on success

### Alternative approaches considered

Considered naming the hook `OnCleanUpHook` (as suggested in the issue), but chose `OnRetriesExhausted` to better describe when the hook is triggered - when retries are actually exhausted, not just for general cleanup.

### Checklist

- [x] I wrote tests for the changes.
- [x] All tests are passing.
- [x] Code has no breaking changes.

```
 message/router/middleware/retry.go      |  7 ++++++
 message/router/middleware/retry_test.go | 43 +++++++++++++++++++++++++++++++++
 2 files changed, 50 insertions(+)
```


zhjdhx:
这是来自QQ邮箱的假期自动回复邮件。 您好，我现在有事，无法亲自回复您的邮件。我将在看到邮件的第一时间给您回复。

m110 (approved):

m110:
Thanks for contributing @Br1an67! I changed the interface a bit to allow extending the args in the future if needed (otherwise we'd need a breaking change to the API). :)

roblaszczak (approved):

PR #672, opened 2026-04-16
fix(example): close HTTP response body and add retry backoff

### LINKED ISSUE #664: run.go: HTTP response body leak and potential goroutine deadlock
The sendCountRequest function in
_examples/real-world-examples/exactly-once-delivery-counter/run.go
contains a critical resource leak that can cause the example to fail under load.

Problem
1. HTTP response bodies are never closed
=> The current implementation calls http.Post in a tight loop but never closes resp.Body:

2. Infinite retry loop without backoff
=> If the server returns any status code other than 204 No Content (e.g., 400 or 500), the function retries immediately in a tight loop

Yanhu007:
Fixes #664

## Problem

`sendCountRequest` in the exactly-once-delivery-counter example:

1. **Never closes `resp.Body`** — leaks file descriptors and TCP connections under load
2. **No retry backoff** — retries immediately in a tight loop on error or non-204 status

## Fix

- Close `resp.Body` after reading the status code
- Add 100ms sleep between retries

### DIFF
--- a/_examples/real-world-examples/exactly-once-delivery-counter/run.go
+++ b/_examples/real-world-examples/exactly-once-delivery-counter/run.go
@@ -139,12 +139,18 @@ func sendCountRequest(counterUUID string) {
 	for {
 		resp, err := http.Post("http://localhost:8080/count/"+counterUUID, "", nil)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


m110:
Addressed in #665, thanks @Yanhu007!

PR #699, opened 2026-09-20
fix(router): return promptly from Close() when handlers were never started

### LINKED ISSUE #697: `Router.Close()` blocks for the full `CloseTimeout` when handlers were registered but the router was never (fully) run
### Steps to reproduce

No infrastructure needed - the bug is entirely in the router's `handlersWg`
accounting, so it can be reproduced with GoChannel.

```go
package main

import (
        "fmt"
        "time"

        "github.com/ThreeDotsLabs/watermill"
        "github.com/ThreeDotsLabs/watermill/pubsub/gochannel"
)

func main() {
        logger := watermill.NewStdLogger(false, false)

        sub := gochannel.NewGoChannel(gochannel.Config{}, logger)

        router, err := message.NewRouter(message.RouterConfig{
                CloseTimeout: 30 * time.Second,
        }, logger)
        if err != nil {
                panic(err)
        }

        router.AddNoPublisherHandler("h", "topic", sub, func(*message.Message) error {
                return nil
        })

        // The router is never Run - e.g. the surrounding service failed later in
        // startup and calls Close() as part of its cleanup.
        start := time.Now()
        _ = router.Close()
        fmt.Println(time.Since(start)) // ~30s, then "router close timeout"
}
```
There are two realistic ways we could get here:

1. Constructed-but-not-run: A service builds a router, registers handlers,
   fails later in startup, and calls `Close()` during cleanup. Close hangs.

2. Partial `RunHandlers` failure: `RunHandlers` registers and starts
   handlers in a loop and returns early if a subscribe fails midway
   (`message/router.go`, line 461):

   ```go
   messages, err := h.subscriber.Subscribe(ctx, h.subscribeTopic)
   if err != nil {
       cancel()
       return errors.Wrapf(err, "cannot subscribe topic %s", ...)
   }
   ```

   `Run` returns that error. Handlers earlier in the map already have their
   goroutines and will call `handlersWg.Done()` as the context unwinds, but any handler the loop
   never reached still sits at `Add(1)` with no goroutine, and a subsequent
   `Close()` waits out the timeout for those.

### Expected behavior

`Close()` on a router with registered but unstarted handlers returns promptly.
An unstarted handler does not increment the wait group, so there is
nothing for `Close` to wait on.

### Actual behavior

`Close()` blocks for the full `CloseTimeout` (30s by default) and then returns
`"router close timeout"`.

`AddHandler` increments `handlersWg` at registration time, but the only matching
`handlersWg.Done()` runs inside the per-handler goroutine that `RunHandlers`
starts. On a router whose handler goroutines never started, `Close()` waits on a
`WaitGroup` that can never reach zero.

In `message/router.go`:

- `AddHandler` registers and counts the handler (line 317):

  ```go
  r.handlersWg.Add(1)
  r.handlers[handlerName] = newHandler
  ```

- The only `handlersWg.Done()` in the package is inside the goroutine that
  `RunHandlers` launches per handler (line 480):

  ```go
  go func() {
      defer cancel()
      ...
      h.run(ctx, middlewares)
      r.handlersWg.Done()   // only Done, only reached once RunHandlers starts the handler
      ...
  }()
  ```

- The only in-package caller of `RunHandlers` is `Run` (line 404). `RunHandlers`
  is exported, so a caller can start handlers added to an already-running router
  itself, but nothing in the package starts them otherwise: no `Run`, no
  goroutine, no `Done`.

- `Close` → `waitForHandlers` waits on that group under the timeout (lines 595-608):

  ```go
  go func() { defer waitGroup.Done(); r.handlersWg.Wait() }()
  ...
  return sync_internal.WaitGroupTimeout(&waitGroup, r.config.CloseTimeout)
  ```

Environment: watermill v1.5.3. Behaviour is transport-independent.

### Possible solutions

Three suggestions, each with pros & cons.

- Count at start, not registration: Move `handlersWg.Add(1)` from
  `AddHandler` into `RunHandlers` beside the `go func()`. Requires synchronisation with Close(). This has a
  complication I hit while trying it: `watchAllHandlersStopped` is started by
  `Run` *before* `RunHandlers` and keys off `handlersWg`, so with the counter no
  longer incremented at registration it observes a transient zero and closes the
  router immediately. On a router with one pre-registered handler, that fires on
  the first iteration of a loop that runs it 300 times. Calling
  `watchAllHandlersStopped` after `RunHandlers` rather than before fixes it,
  though the `hasNoHandlersYet` sampling of `len(r.handlers)` is racy (this is a pre-existing condition) for
  handlers that start and finish before the watcher arms.

- Have `Close` skip or settle unstarted handlers: Complication: lock
  ordering — `Close` takes `closedLock` then `handlersLock`, `RunHandlers` takes
  `handlersLock` alone, so racing a `Close` against a `RunHandlers` risks a
  double `Done`.

Happy to open a PR for whichever shape you'd prefer.

PS thank you for the amazing content on your website, it has been so useful for me.

baltasarblanco:
Fixes #697.

### Motivation / Background

`Router.Close()` blocks for the full `CloseTimeout` and returns `router close timeout` when handlers were registered but never started.

`AddHandler` increments `handlersWg` (`message/router.go:317`), but the matching `Done()` only runs inside the goroutine that `RunHandlers` starts (`message/router.go:480`). A handler that is registered and never started therefore leaves the counter incremented with nothing behind it, and `waitForHandlers()` waits it out.

This is a regression from #266 (241bc78, "Allowing to add handlers while router is already running"), which moved `handlersWg.Add(1)` out of the `RunHandlers` loop and into `AddHandler`. Running the reporter's reproducer against both trees with `CloseTimeout` set to 3s:

```
241bc78^   Close() took 0s,  err=<nil>
v1.5.3     Close() took 3s,  err=router close timeout
```

There are three ways to reach it:

1. **Constructed but not run.** A service builds a router, registers handlers, fails later in startup, and calls `Close()` during cleanup.
2. **Partial `RunHandlers` failure.** `RunHandlers` starts handlers in a loop and returns early when a subscription fails (`message/router.go:459-462`). Handlers the loop already reached stop normally as the context unwinds; handlers it never reached are still counted.
3. **Handler added to a running router.** Handlers added after `Run()` are only started by an explicit `RunHandlers()` call (see the note on `AddHandler`, `message/router.go:271`). Between `AddHandler` and `RunHandlers` the counter is already incremented with no goroutine behind it, and that window is controlled by the caller.

### Detail

For every handler that was never started, `Close()` now does what the handler's own goroutine would have done on a router-wide close: it closes the handler's subscriber and publisher, releases its wait group counter, and drops it from `r.handlers` the same way a normally-stopped handler is dropped (`message/router.go:483-485`). A started handler closes its subscriber in `handleClose` (`message/router.go:800`) and its publisher at the end of `run` (`message/router.go:671`); an unstarted one previously did neither. Closing them runs in a goroutine that calls `handlersWg.Done()` when finished, so `waitForHandlers` bounds it by `CloseTimeout` exactly as it does for started handlers. Thanks to @markgoddard for spotting the unclosed subscriber and publisher in review and for the change that closes them.

`Close()` already holds `r.handlersLock` for its whole body (`message/router.go:567`), the same lock `RunHandlers` holds while it sets `h.started`, so the sweep cannot observe a handler mid-start. Removing swept handlers from `r.handlers` is what rules out the double `Done()` raised in the issue: a later `RunHandlers()` only walks the map (`message/router.go:436`), so a handler released here can no longer be started and decremented a second time. The closing goroutine does not take `handlersLock`. The normal path, where every handler is started and decrements the counter itself, is untouched.

Behaviour before and after, with `CloseTimeout` at 3s:

| scenario | before | after |
| --- | --- | --- |
| handler registered, router never run | 3s, `router close timeout` | 0s, nil |
| router with no handlers | 0s, nil | 0s, nil |
| `Run()` with a subscriber that fails partway | 3s, `router close timeout` | 0s, nil |
| happy path: `Run`, process a message, `Close` | 0s, nil, message delivered | 0s, nil, message delivered |
| handler added while running, `RunHandlers` not called | 3s, `router close timeout` | 0s, nil |

Three regression tests cover paths 1, 2 and 3, and a fourth checks that the subscriber and publisher of an unstarted handler are closed. They assert on the returned error and on recorded state rather than on wall-clock time, and each is guarded by a `select` that fails well before `CloseTimeout` so a regression cannot hang CI. All four fail without the corresponding change and pass with it.

**Moving `handlersWg.Add(1)` back into `RunHandlers`**, so that unstarted handlers never increment the counter at all. This is the more obvious reading of the bug, but it does not work:

- It is precisely what #266 changed, and undoing it takes the dynamic-handler support with it. Path 3 above reopens: a handler added to a running router would no longer be counted until `RunHandlers` is called, so if the previously running handlers stop inside that window, `watchAllHandlersStopped` closes the router.
- `Run()` starts `watchAllHandlersStopped` before `RunHandlers` (`message/router.go:402-404`). With `Add(1)` in `RunHandlers`, the watcher goroutine can reach `handlersWg.Wait()` before the handlers are counted, observe zero, and close the router during startup.

I tried it to be sure. `TestRouter_functional`, ten runs, each in its own process:

```
run  1: PASS      ("All handlers stopped" x0)
run  2: FAIL      ("All handlers stopped" x1)
run  3: PASS      ("All handlers stopped" x0)
run  4: HUNG      ("All handlers stopped" x1)
run  5: HUNG      ("All handlers stopped" x1)
run  6: HUNG      ("All handlers stopped" x1)
run  7: PASS      ("All handlers stopped" x0)
run  8: FAIL      ("All handlers stopped" x1)
run  9: FAIL      ("All handlers stopped" x1)
run 10: FAIL      ("All handlers stopped" x1)
```

Seven of ten runs broken: four failed with dropped messages, three hung until the test binary hit its timeout. `All handlers stopped, closing router` is logged in exactly those seven and in none of the three that passed. `TestRouter_functional_nack` fails the same way.

**Also closing `h.stopped` for swept handlers.** `h.stopped` is only created in `RunHandlers` (`message/router.go:469`), so for a handler that never started it stays nil and `Handler.Stopped()` (`message/router.go:718`) returns a nil channel that blocks forever. That is an adjacent defect rather than this one, so I left it out to keep the change minimal — happy to fold it in here if you'd prefer.

### Checklist

- [x] I wrote tests for the changes.
  - `TestRouter_close_when_handlers_were_not_started`
  - `TestRouter_close_when_running_handlers_failed`
  - `TestRouter_close_when_handler_added_while_running_was_not_started`
  - `TestRouter_close_closes_pubsub_of_handlers_that_were_not_started`
  - Each fails without the corresponding change and passes with it.
- [x] All tests are passing.
  - `make test_short`, `make test`, and `make test_race` all pass (Go 1.27.1, Linux amd64).
  - `go test ./message/... -count=10 -race` passes with no data races.
  - `gofmt -l message/` is clean, `go vet ./message/` is clean, and `golangci-lint run` (v2.13.2) reports 0 issues.
- [x] Code has no breaking changes.
  - No exported API changes. `Close()` no longer waits for handlers that have no goroutine to wait for, and it closes their subscriber and publisher, as it already does for started handlers.
- [x] _(If applicable)_ documentation on [watermill.io](https://watermill.io/) is updated.
  - Not applicable: no documented behaviour changes.


markgoddard (commented):
Thank you for picking this up @baltasarblanco.

I tried the other approach (`wg.Add()` immediately before `go`), which to me seems a more typical use of wait groups. However, there were several knock on effects that needed to be worked out.

This looks like a good pragmatic solution. I reviewed it myself and couldn't see any issues. I pointed Claude at the change and it highlighted that the handler's subscriber and publisher are not closed for unstarted handlers in this solution. It came up with the attached diff as a solution.

[close-pub-sub.patch](https://github.com/user-attachments/files/32465708/close-pub-sub.patch)

baltasarblanco:
Thanks @markgoddard, good catch. A started handler closes its subscriber in `handleClose` and its publisher at the end of `run` on a router-wide close, and an unstarted one did neither, so this brings the two paths in line.

I've applied your change as-is, with you as co-author. The new test fails without the `router.go` part and passes with it, and the full suite, `-count=10 -race`, `go vet` and `golangci-lint` are clean. I've updated the description to match.

